PR #615, opened 2015-07-29
Remove silent error consumption in getStringValue(...)

sprocter:
The ```getStringValue``` method in the PropertyUtils class fails silently, instead of throwing any exceptions that get raised. This fix resolves that.


reteprelief:
Hi Sam,

we updated the method getStringValue and pushed it in the develop branch. It now passes on the exception. We have introduced a second method that takes an additional parameter to provide a default value - as is the case for other get methods, e.g., real, int.

PR #804, opened 2017-03-22
Remove unnecessary self dependency in org.osate.aadl2.edit

smithdtyler:
This self dependency breaks Eclipse dependency analysis tools.


lwrage:
Thanks! BTW, I looked into this yesterday and found that it's the EMF code generation that adds these dependencies.

Issue #613, opened 2015-06-25
Error markers of non xtext annexes interfere with expansion of model files in navigator


lwrage:
Somehow markers are created and immediately deleted again. It seems to happen when the navigator loads the changed file. As an example take the model from #611 and edit it to fix/introduce errors. Exception about missing markers are created in the error log whenever the file is saved.

lwrage:
This doesn't happen any more. Closing.

Issue #587, opened 2015-05-08
ResourceException in AadlNavigator


joeseibel:
The exception happens when I move an aadl file from one directory to another within the same project by dragging and dropping.

Stack trace:

```
org.eclipse.core.internal.resources.ResourceException: The resource tree is locked for modifications.
    at org.eclipse.core.internal.resources.WorkManager.checkIn(WorkManager.java:116)
    at org.eclipse.core.internal.resources.Workspace.prepareOperation(Workspace.java:2251)
    at org.eclipse.core.internal.resources.Resource.deleteMarkers(Resource.java:844)
    at org.osate.aadl2.modelsupport.errorreporting.MarkerParseErrorReporter.deleteMessagesImpl(MarkerParseErrorReporter.java:117)
    at org.osate.aadl2.modelsupport.errorreporting.AbstractErrorReporter.deleteMessages(AbstractErrorReporter.java:56)
    at org.osate.aadl2.modelsupport.errorreporting.ParseErrorReporterManager.getReporter(ParseErrorReporterManager.java:116)
    at org.osate.xtext.aadl2.parsing.AnnexParserAgent.afterModelLinked(AnnexParserAgent.java:122)
    at org.eclipse.xtext.linking.impl.AbstractCleaningLinker.linkModel(AbstractCleaningLinker.java:53)
    at org.eclipse.xtext.resource.XtextResource.doLinking(XtextResource.java:290)
    at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doLinking(LazyLinkingResource.java:107)
    at org.eclipse.xtext.resource.XtextResource.updateInternalState(XtextResource.java:258)
    at org.eclipse.xtext.resource.XtextResource.updateInternalState(XtextResource.java:248)
    at org.eclipse.xtext.resource.XtextResource.doLoad(XtextResource.java:161)
    at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doLoad(LazyLinkingResource.java:100)
    at org.eclipse.emf.ecore.resource.impl.ResourceImpl.load(ResourceImpl.java:1518)
    at org.eclipse.emf.ecore.resource.impl.ResourceImpl.load(ResourceImpl.java:1297)
    at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.demandLoad(ResourceSetImpl.java:259)
    at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.demandLoadHelper(ResourceSetImpl.java:274)
[...]

PR #906, opened 2017-09-27
Keep removal of markers in the caller's thread 

smithdtyler:
This change appears to avoid race conditions in creation/deletion of markers, specifically issue https://github.com/osate/osate2-core/issues/891


lwrage:
Thanks! I'll see if this also fixes #613.

philip-alldredge:
Why trying to track down a deadlock, I came across this as being involved. I'm looking into the cause on the graphical editor side. However, This change reintroduces issue #587. The commit log indicates the usage of a job was intended to fix that issue.

smithdtyler:
This strikes me as an operating ordering problem. Is there something equivalent to Swing's Event Dispatch Thread for handling marker creation and destruction?

philip-alldredge:
I believe the deadlock in the graphical editor has been fixed. As far as the problem with creating/deleting markers and #587.. There are mechanisms similar to Swing's event dispatch depending on whether it should be be performed in the main thread or using job mechanism but I don't know enough about the situation to comment on the proper behavior.

PR #925, opened 2017-10-23
Update ELK version to 0.3

philip-alldredge:


philip-alldredge:
@lwrage this update the version of ELK used in the setup file.  I was waiting on this change but would like to get this merged to avoid merge conflicts anytime the setup file changes.

lwrage:
@philip-alldredge Are you sure this is correct? I now have a development environment that has elk 0.2.2 installed, but the running OSATE uses 0.3.0.

philip-alldredge:
Yes, I believe so. The ELK features don't need to be installed in the development environment. Unfortunately, it doesn't remove them from your development environment after they are removed from the setup. 

Sometimes it is helpful to have additional ELK tools installed while development but those need to be installed in the runtime environment during development but which are better left out of the normal runtime environment.  I was originally manually adjusting the target during development to add some of the plugins to the runtime environment.  This is tedious and caused problems. 

The osate-ge branch related to ELK has a setup file that can be used temporarily when doing such work: https://github.com/osate/osate-ge/blob/elk_integration/org.osate.ge/setups/layout_dev.setup 

This prevents having to manually adjust targets and keeps the dependencies out of normal run-time target.

lwrage:
You can add the development tools using plain P2 director task. Then they will not be available in the running OSATE. I think that all useful development tools should be part of the main setup. Otherwise we end up with additional instructions on how to set up a development eclipse that just apply to the graphical editor.

philip-alldredge:
These ELK tools need to be installed in the running Eclipse. They include a view in the running Eclipse of the last graph that was processed. I suppose some of them could be used and added to the development environment but that wouldn't reduce the need for having them in the runtime environment. Would like me to include them in the development environment?

lwrage:
If there's any ELK tool that you want to use during development I suggest that you add it to the setup.
I just want to make sure that there's a single setup that installs all the needed/useful tools for OSATE development.

philip-alldredge:
See #927

PR #945, opened 2017-11-21
Fix for instantiate model action

### LINKED ISSUE #943: Instantiate action active on interface components
The "Instantiate" action is active in the Outline view context menu for elements that aren't system implementations.  Invoking the action on these items results in an exception:

	at org.osate.ui.handlers.InstantiateHandler$1.lambda$0(InstantiateHandler.java:168)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:119)

AaronGreenhouse:
Fixes #943 

Action and menu command updated so that the instatiate model action is correct.


AaronGreenhouse:
Okay, I reverted InstanceGrammarAccess, not sure what happened there.

PR #966, opened 2017-12-07
New file wizards

### LINKED ISSUE #946: Complete set of new file wizards
We need new file wizards for all of the Alisa file types. Also, the existing package and property set wizards should be added to the New menu in the Main Menu and the AADL Navigator Popup.

New file wizards can be added to the New menu in the Main Menu with the class `AadlPerspectiveFactory` or the extension point `org.eclipse.ui.perspectiveExtensions`. They can added to the AADL Navigator with the extension point `org.eclipse.ui.navigator.navigatorContent`.

The `NewModelWizard` should also be split into separate wizards for packages and property sets.

joeseibel:
Fixes #946.

Should be merged alongside osate/alisa#35.


org.osate.ui/src/org/osate/ui/wizards/NewModelWizard.java
```diff
@@ -95,7 +95,7 @@
  * @author jseibel
  */
 abstract class NewModelWizard extends Wizard implements INewWizard {
-	private static final String[] HIDE_FOLDERS = { "instances", "diagrams", "imv" };
+	private static final String[] HIDE_FOLDERS = { ".aadlbin-gen", ".settings", "instances", "diagrams", "imv" };
```
lwrage: Should hide all directories that start with '.'

joeseibel: Done.

PR #973, opened 2017-12-13
AADL project wizard

### LINKED ISSUE #729: AADL project wizard improvements
The AADL project wizard should no longer create the .aadlsettings file.
.aadlbin-gen should be automatically added to a .gitignore file

joeseibel:
Fixes #729.

### DIFF
--- a/org.osate.ui/src/org/osate/ui/wizards/AadlProjectWizard.java
+++ b/org.osate.ui/src/org/osate/ui/wizards/AadlProjectWizard.java
@@ -36,10 +36,11 @@
  */
 package org.osate.ui.wizards;
 
-import java.io.IOException;
+import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


org.osate.ui/src/org/osate/ui/wizards/AadlProjectWizard.java
```diff
@@ -320,17 +308,23 @@ public boolean performFinish() {
 		selectAndReveal(newProject);
 		final IProject p = getNewProject();
 
-		String filepath = p.getFile(WorkspacePlugin.AADLPATH_FILENAME).getRawLocation().toString();
```
lwrage: Please remove this and the following constants from WorkspacePlugin. All of them shouldn't be used anywhere else. If they are, that's wrong.

PR #983, opened 2017-12-20
Assert scope filter

### LINKED ISSUE #982: OsateTest.assertScope filters too much
The `assertScope` will filter out contributed model unit names even when scoping for a local name. To reproduce, create a contributed property set with the name `m1`. The test `ModeTransitionsTest` will then fail because `m1` is filtered out of the scope.

joeseibel:
Fixes #982.


joeseibel:
The problem with `EventInitiatorTest` is fixed with osate/ErrorModelV2#163.

PR #985, opened 2017-12-21
Fixes #961: Fixed instantiateSubcomponent()

### LINKED ISSUE #961: Instance model with abstract instance instead of process
The instance model contains an abstract component instance pointing to the process subcomponent, whose classifier is abstract. Should be process.

package abstractprocess
public
	system s
	end s;
	system implementation s.i
		subcomponents
			p: process a;
	end s.i;
	abstract a
		end a;
end abstractprocess;

AaronGreenhouse:
I think the category can be determined as follows:

If the classifier is not abstract then use its category (and maybe check
that the subcomponent either has the same category or is abstract).
If the classifier is abstract then use the category from the
subcomponent.

Only if both are abstract the component instance should be abstract.
If both are not abstract then they must have the same category. If the
categories are different, validation should already have reported an
error, and we don't instantiate models with errors. It can't hurt if the
instantiator checks again, though.


AaronGreenhouse:
Created the tests and pushed back out to the branch

PR #1023, opened 2018-01-21
1009 unresolved names

### LINKED ISSUE #1009: Name resolution issues
The capstone model templates show bogus errors after clean all in OSATE 2.3.1. Everything is fine in 2.2.3.
The errors go away if a file with errors is opened in the text editor, trivially modified, and saved.

lwrage:
fixes #1009


lwrage:
This works for the simplified test case but not for the larger model. It still has errors in refinement 4.

PR #1034, opened 2018-02-06
Enable formatting by default.

### LINKED ISSUE #1026: Inverse Feature Group Source Format
Using the graphical editor to set a Feature Group to inverse and then non-inverse creates a missing space in the source.

Right click on Feature Group -> Set Inverse.
Right click on Feature Group -> Set Non-Inverse.

Result in source: 
```
fg: feature groupfg_classifier;

joeseibel:
Fixes #1026.


lwrage:
@philip-alldredge The formatter was not used by the serializer, the fix turns formatting on. Could you test this for a couple of days, please? It fixes the immediate issue, but I want to make sure that it doesn't cause new issues before merging the pull request.

philip-alldredge:
@lwrage @joeseibel Unfortunately, I was able to reproduce the issue with this branch.
The change fixes the issue if the text editor is open but when the text editor is not open, the feature group is not serialized appropriately. The difference for the graphical editor is that when the Xtext editor is not open, it opens the AADL model as an EMF resource directly and modifies the model using setInverse().  It is strange that this is misbehaving only affecting feature groups.

For future reference, should I respond to the review or the issue in cases such as this?

philip-alldredge:
Added a easier way to reproduce the issue to #1026.

PR #1035, opened 2018-02-07
Fixed 1029 - merged the textual and graphical package wizards

### LINKED ISSUE #1029: Provide single wizard for new AADL package
Currently there are 2 wizards to create a new package (textual vs. graphical). It would be better to have a single wizard with two ok buttons or a checkbox to invoke either the textual or graphical editor.
I propose to use the wizard for textual AADL packages and extend it to call the graphical editor. This wizard allows selection of a directory for the package, whereas the graphical editor puts all packages in a directory named 'packages.'

AaronGreenhouse:
Fixed #1029.


philip-alldredge:
This should be merged in conjunction with the branch in the GE repo. I pushed an additional change there to remove the file for the old wizard.

philip-alldredge:
@lwrage @AaronGreenhouse As a result of the work being done in osate/osate-ge#211, the combined wizard will open a "Create Diagram" dialog when the user selects to open in the Graphical Editor. This dialog is being introduced to prompt the user for a diagram name and a diagram type when a new diagram is created. I personally think this is reasonable behavior and would give the user a chance to select the name for the diagram. Another option is that it could create a Package Diagram with a default name. Thoughts?

lwrage:
I'd be fine with the additional dialog if the diagram name is pre-populated
with the package name and a sensible default for the diagram type, so it
would usually just require one additional mouse click.

philip-alldredge:
Sounds good.

PR #1045, opened 2018-02-09
Fixed 831 by adding resource change listener to OsateCorePlugin

### LINKED ISSUE #831: Lost reference to renamed project
If one project is dependent on another.  If the project that is being dependent upon is renamed, the reference is lost.

AaronGreenhouse:
Fixed #831.  Tested as explained in #831.

### DIFF
--- a/org.osate.core/src/org/osate/core/OsateCorePlugin.java
+++ b/org.osate.core/src/org/osate/core/OsateCorePlugin.java
@@ -43,6 +43,9 @@ import java.util.MissingResourceException;
 import java.util.ResourceBundle;
 
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
+import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.ResourcesPlugin;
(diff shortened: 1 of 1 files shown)


AaronGreenhouse:
I htought I did that already.  I'll check on it.

AaronGreenhouse:
I was using Workspace.run() inside my Job, but I guess that isn't quiet the same thing.

AaronGreenhouse:
My RenameJob in OsateCorePlugin uses Workspace.run() and a IWorkspaceRunnable internally.  Workspace.run() already batches all the change events, which is why I did this.

The WorkspaceModifyOperation is also based around using Workspace.run(), but is more geared towards UI-specific operatoins.  In particular, I don't have any way of actually running a WorkspaceModifyOperation, because my event handler odesn't have access to an IRunnableContext as would be available from insider a dialog box or wizard or some such thing like that.

lwrage:
You can also just call `op.run(aMonitor)` to execute a `WorkspaceModifyOperation op`. This is what we do in class OsateText As you already take care of the batching, it wouldn't make a difference in functionality, but IMHO the code would be a little clearer.

AaronGreenhouse:
But the WMO is still going to have to be in a separate job because it's being invoked by a resource listener that is already executing in the workspace.  I'm not sure that is going to make things any clearer.

It works the way it is now.

lwrage:
Ah, my bad. I thought the WMO would create a job when it's run.

PR #1064, opened 2018-02-20
Fixes #1063.

### LINKED ISSUE #1063: Aadl2SemanticSequencer Should Use Editing Domain of Resource Set
An issue with how the Aadl2SemanticSequencer handles annexes in related to the graphical editor has been found. The issue is that the graphical editor uses its own resource set and editing domain when performing changes. The reason for this is that it uses live resources and often manipulates the contents of the resource set and it is intended to avoid conflicts with other users of the resource set. When an xtext document is open, an editing domain should not be used because then xtext doesn't use editing domains.

However, the Aadl2SemanticSequencer uses the registered OSATE editing domain in all cases. If editing an xtext document then this doesn't cause an exception because a transaction is not required. However, if the document is not opened then not using the graphical editor's editing domain results in an exception due to a transaction not being active. 

The recommended solution is to use the editing domain associated with the semantic object. If an editing domain does not exist, then one should not be used.

philip-alldredge:
Fixes #1063. Adjusts Aadl2SemanticSequencer to use the editing domain of
the resource set.

### DIFF
--- a/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/serializer/Aadl2SemanticSequencer.xtend
+++ b/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/serializer/Aadl2SemanticSequencer.xtend
@@ -69,10 +69,9 @@ class Aadl2SemanticSequencer extends AbstractAadl2SemanticSequencer {
 				// otherwise use the original annex text
 				if (parsedLibrary !== null && annexUnparser !== null) {
 					try {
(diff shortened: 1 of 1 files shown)


philip-alldredge:
I'd like to clarify the reasoning for moving the unparsing into the command. It is needed because there are annex unparsers such as the EMF that set the name of the clause in order to fix serialization issues. If it is not in the command, then a transaction error will occur if that resource set has a transacitional editing domain.

PR #1066, opened 2018-02-21
Updated table of content to refer to "Analyses" instead of "Plugins".

reteprelief:
Fixes osate/osate2-core#1065
Removed outline subentries under plugins (analyses) to out of date
documentation.

Updated index.html of plugins (analyses) to refer to most up to date
documentation on latency and safety analysis. It also has the other
recently updated documentation that most people missed as it was only
available on the plugins outline entry.

Note: I copied the latency and safety documentation from their original location to avoid cross plugin references. The filed copied are latency.html/md safetyanalysis.html/md and safetyanalysisexample.html/md as well as the images folder.


lwrage:
Does this fix 1048 and 1049, and 1065 is a duplicate of these?

reteprelief:
Lutz,

In 1048 and 1049 I would expect the outline of each of the documents – at least the first or second level -  to be reflected in TOC. You said someone has a script to auto-generate this.
At the moment I just hooked in the whole document so it is available.

Peter

PR #1128, opened 2018-04-11
Fixed #1092

### LINKED ISSUE #1092: Instantiation fails when referencing a required mode in a property association.
In the following AADL model, OSATE fails instantiating the system implementation main.impl
```
package required_modes_instantiation
public
	system main
	end main;
	
	system implementation main.impl
	subcomponents
		proc: process proc.impl;
	properties
		Period => 1000 ms in modes (m1), 200 ms in modes (m2) applies to proc.t;	
	end main.impl;
	
	process proc
		modes
			m1: initial mode;
			m2: mode;
	end proc;
	
	process implementation proc.impl
	subcomponents
		t: thread t;	
	end proc.impl;
	
	thread t
		requires modes
			m1: initial mode;
			m2: mode;
	end t;
	
end required_modes_instantiation;
```
The reason seems to be that I reference required modes (m1 and m2) in the property association of property Period for thread subcomponent t. The property association is given in system implementation main.impl.

If I remove the keyword "requires" in thread type declaration t, the instantiation works fine.

Let me know if this is a bug or a miss-use of the language.
Best regards,
Etienne.

AaronGreenhouse:
Fixed this.  Added Junit tests.


AaronGreenhouse:
Checked in one change to my code that to fix the Serializer2Test.testParentMode test in
org.osate2.instance.textual.tests.  So the test found an actual bug.

The other broken test Serializer2Test.testModes in org.osate2.instance.textual.tests is broken because it needs to be updated.

AaronGreenhouse:
Fixed the test case  Serializer2Test.testModes in org.osate2.instance.textual.tests  to reflect the new reality after changes in this issue.

PR #1153, opened 2018-04-19
Fixes null pointer in get properties of ARINC schedule

reteprelief:
Fixes osate/osate2-core#1126
Get properties of ARINC653 schedule did not handle incomplete record fields correctly.


reteprelief:
reverted the commit. Redid the correction for the null pointers. Hopefully this should have removed the deletion of .project from this pull request. btw this was for the resolute.help plugin when it was in osate2-core. We deleted it when Lutz was with me.

PR #1258, opened 2018-05-30
Process annexes without depending on the Eclipse UI

### LINKED ISSUE #1257: Annex processing assumes that it runs inside Eclipse
Annex parsing is implemented in a rather weird way that requires it to run in Eclipse. As a consequence, we cannot parse AADL files standalone, for example, in tests. This needs to be redesigned from scratch. Maybe #619 is the way to go.

child of #1018
blocks #1254

lwrage:
fixes #1257 

Eliminated use of marker error reporter for parsing and linking.
Made error reporters independent of IResource.


lwrage:
I think you should have requested changes, instead of approving the PR.

PR #1263, opened 2018-05-30
1261 latency doc

### LINKED ISSUE #1261: Latency analysis result contains unnecessary data and description is not a section under Latency Analysis
The latency analysis result contains latency contributor data that are intermediate compute results but are not included in the report. Therefore they are not useful to be recorded in AnalysisResult.

The description is currently added into the TOC parallel to latency analysis instead of a section within latency analysis.

reteprelief:
Fixes #1261


reteprelief:
When I moved the documentation I noticed that I had included temporary computations on latency contributors in the Result objects. 
The code change and documentation update were part of a single commit. Instead of manually undoing the changes and then redoing them I updated the issue description.

lwrage:
Please try to keep things separate for future PRs. We really need to become more disciplined about this.

PR #1282, opened 2018-05-31
Add a method to instantiate without saving to a file

### LINKED ISSUE #1255: Instantiation always creates error markers
There should be a way to invoke instantiation without creating error markers on the instance model file.

child of #1018
needed by #603

lwrage:
fixes #1255
Errors and warnings are added to a QueuingAnalysisErrorReporter.


lwrage:
build!

PR #1292, opened 2018-06-01
Use shared resource set for tests

### LINKED ISSUE #1254: Tests should not use a workspace build
Currently, tests depend on running in a workspace. We create files in the workspace and do a build, which takes about 4s per test.

child of #1018
blocked by #1255

lwrage:
With these changes tests can use a shared resource set instead of a workspace build. Contributed AADL files are loaded only once into the resource set.
fixes #1254


core/org.osate.aadl2/src/org/osate/aadl2/util/Aadl2Util.java
```diff
@@ -128,9 +128,9 @@ public static NamedElement findOwnedNamedElement(Classifier owner, String name)
 	 */
 	public static Property lookupPropertyDefinition(final EObject context, final String propSetName,
 			final String propName) {
-		if (propertyService != null) {
-			return propertyService.lookupPropertyDefinition(context, propSetName, propName);
-		}
+//		if (propertyService != null) {
```
joeseibel: Why are you leaving commented out code in the repo?

PR #1295, opened 2018-06-05
Added version requirement for ELK dependencies.

### LINKED ISSUE #1244: GE Doesn't Specify ELK Version Dependency
The Graphical Editor doesn't define which version of the ELK it requires. For pre-built OSATE products, this does not cause an issue. However, when installing into an existing Eclipse installation, an incorrect version of ELK could be used. Need to add the version to the manifest.

philip-alldredge:
Fixes #1244.

### DIFF
--- a/core/org.osate.build.product/osate.product
+++ b/core/org.osate.build.product/osate.product
@@ -74,6 +74,7 @@
       <repository location="http://osate-build.sei.cmu.edu/download/osate/testing/updates/" enabled="true" />
       <repository location="http://download.eclipse.org/eclipse/updates/4.7" enabled="true" />
       <repository location="http://download.eclipse.org/releases/oxygen" enabled="true" />
+      <repository location="http://download.eclipse.org/elk/updates/releases/0.3.0/" enabled="true" />
    </repositories>
 
 </product>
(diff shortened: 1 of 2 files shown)


philip-alldredge:
I've added the update site to the OSATE product.

lwrage:
Sorry, I was confused. The elk features/plugins should be added to the osate p2 repo.

philip-alldredge:
Okay, I've reverted the change to the OSATE product. I don't know anything about the p2 repo. What needs to be done? The ELK repo URL is http://download.eclipse.org/elk/updates/releases/0.3.0/

lwrage:
I _think_ the elk features that are in the target platform file must be added to the product file. That should add them to the repository.

philip-alldredge:
I've added the features to the product.

lwrage:
There is a way to put a reference to another repository (B) into a p2 repository (A). When installing from A, B is added to the list of available repositories for installation. Unfortunately, these references are dropped during the tycho build (https://bugs.eclipse.org/bugs/show_bug.cgi?id=453708).

philip-alldredge:
So, what needs to be changed?

PR #1304, opened 2018-06-08
#1301 Use qualified Resolute claim or compute function name in verify method registry

### LINKED ISSUE #1301: Verify method registry uses identifier to refer to Resolute function rather than qualified name
The method registry could use a qualified name, which would get resolved by the global scope provided used by Verify.
The change can be made backward compatible.
Once all method registries use qualified references, we can remove the VierfyLinkingService.

reteprelief:
Fixes #1301


alisa/org.osate.verify/src/org/osate/verify/linking/VerifyLinkingService.java
```diff
@@ -1,50 +0,0 @@
...
-import org.eclipse.emf.ecore.EReference;
-import org.eclipse.xtext.linking.impl.DefaultLinkingService;
-import org.eclipse.xtext.linking.impl.IllegalNodeException;
-import org.eclipse.xtext.nodemodel.INode;
-import org.eclipse.xtext.resource.IEObjectDescription;
-import org.osate.verify.verify.ResoluteMethod;
-
-import com.rockwellcollins.atc.resolute.resolute.ResolutePackage;
-
-public class VerifyLinkingService extends DefaultLinkingService {// PropertiesLinkingService {
```
lwrage: Please mark as deprecated and add a deprecation comment with a target version for removal. Otherwise we may forget to remove it.

reteprelief:
Added deprecated to VerifyLinkingService

PR #1308, opened 2018-06-08
Graphical Editor Memory Leak Fixes

### LINKED ISSUE #1307: Fix Memory Leaks in the Graphical Editor
There are global references to the graphical editor once an editor is closed. This prevents the objects from being garbage collected.

- [x] AgeDiagramBehavior - Post Selection Listener 
- [x] AgeDiagramBehavior - Part Listener

There appears to be some other interaction with Eclipse that is occurring that is causing the editor to be retained until a new editor, textual or graphical, is opened. This latter issue will be ignored for now as it is uncertain whether that is normal Eclipse behavior.

philip-alldredge:
Resolves multiple causes of objects not beeing garbage collected.


RyanMcilnay:
GraphitiEditPartAdapterFactory was throwing a null exception on close of editor because graphitiAgeDiagram was now being set to null in AgeDiagramBehavior during dispose.  I believe it only occurred when a save was needed on close.

I added a null check, but I'll hold off on merging for now til we discuss it.

Issue #1182, opened 2018-04-25
Improve Undo and Redo Support


philip-alldredge:
> support for undoing and redoing actions. 

Undo and redo support is currently very limited. It needs to be expanded to fix issues with the dirty state indicator and to allow undoing any operation as long the text editor isn't being used. If the text editor is being used, then undo support will be reduced.

- [x] An undo stack which is shared across editors should be implemented.
- [x] Undoing or redoing an action will result in the appropriate editor being activated.
- [x] Changes to diagram configuration should be undoable.
- [x] The undo stack should include diagram changes and model changes from the graphical editor including the properties view. Model changes along with any associated diagram changes should be represented as a single entry on the undo stack.
- [x] Updates that do not affect any diagram and were not caused by the graphical editor should not be shown in the undo stack. 
- [x] Model updates outside the graphical editor should reset the action stack. This includes updates caused by opening and closing the AADL source.
- [x] Renaming should be undoable.
- [x] Diagram should be dirty if and only if there was a visual change to the diagram itself.
- [x] Undoing an operation that caused the diagram to be dirty should result in the diagram no longer being dirty.
	
Also include #1199

philip-alldredge:
Rationale:
### Global Undo
A shared undo/redo strategy was chosen because multiple diagrams may be interacting with the same AADL packages.  In such cases, actions must be reversed in the order they were performed.

### Text Editor Limitations
Undo and redo behavior will be degraded when the text editor is open due to implementation challenges. The text documents have their own undo stacks. When the graphical editor edits such a model, the change is added to the document's stack.  
While invalidating the undo stack when the text editor is opened or used is sub-optimal, it is imagined that in most cases a user will want to undo changes immediately after performing them rather than after opening the text editor.  With the text editor open the behavior will be similar to the current behavior but when it is closed, the undo behavior will be significantly enhanced by allowing reversing of model changes.

philip-alldredge:
ebd75e16291c96a3346660cbc35009c94f989b5e

PR #1328, opened 2018-06-15
GE undo redo support

philip-alldredge:
Implements #1182


philip-alldredge:
Looks like something happened when merging develop into the branch at some point. There are lots of unintended whitespace/line ending file changes. Must be fixed before merged. Changes should be limited to GE plugins.

Issue #1246, opened 2018-05-16
Diagrams are automaticly marked as changed


stevevestal:
## Summary
When I save a diagram, a few seconds later the Zoom Out icon flickers a few times, and the diagram is marked dirty again.  I do not see any visual changes in the diagram itself.  Selecting save causes this behavior to repeat.

## Steps to Reproduce
This always happens for me, but I suspect it might have to do with my configuration.  I am running this as an Eclipse Application from within Eclipse RCP.  (I would rank this low priority/ a curiosity unless it is happening in more typical user configurations.)

## Environment

* **OSATE Version**: OSATE nightly test installed into Eclipse Oxygen RCP with ELK 0.3.0.
* **Operating System**: Windows 10

stevevestal:
Not all diagrams.  Only some.  The only difference I can see is that the ones that auto-dirty are too large (at the default 100%) to fit in any single screen.  Changing the default zoom, or shrinking the window size for the diagram that does not do this, doesn't make any difference.

philip-alldredge:
This should be investigated when looking at similar issues. See #1199.  The more information the better but when we rework how the dirty detection is determined, this issue should be fixed.

philip-alldredge:
Fixed in ebd75e16291c96a3346660cbc35009c94f989b5e

Issue #1199, opened 2018-04-25
Dirty marker doesn't disappear when undo-ing all changes in graphical editor


philip-alldredge:
<a href="https://github.com/lwrage"><img src="https://avatars1.githubusercontent.com/u/750311?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [lwrage](https://github.com/lwrage)**
_Wednesday Jan 31, 2018 at 16:52 GMT_
_Originally opened as https://github.com/osate/osate-ge/issues/283_

----

Open a diagram, resize a component, hit undo. The diagram looks the same as before but the diagram is still marked as dirty.
Maybe a check of the undo stack needs to be added?

philip-alldredge:
<a href="https://github.com/philip-alldredge"><img src="https://avatars2.githubusercontent.com/u/1376364?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [philip-alldredge](https://github.com/philip-alldredge)**
_Wednesday Jan 31, 2018 at 17:01 GMT_

----

This will need to be addressed when the Undo and Redo support is reworked. The current dirty indicator just sets the flag when something is changed. This resolved the previous issue but does not result in the flag being cleared as part of the undo.

Getting useful information from the undo stack is tricky because it includes things that don't result in a diagram change. This was the root cause of the dirty indicator constantly being reset. 

This needs to be worked in conjunction with old_216

philip-alldredge:
Fixed in ebd75e16291c96a3346660cbc35009c94f989b5e

PR #1335, opened 2018-06-19
Undo/Redo rework.

philip-alldredge:
Implements #1182, #1199, and #1246.

Replaces the current undo capability that is is
very limited with an implementation that allow undoing and redoing a
wider range of actions. Usage of the text editor can degrade
capabilities.

- Removed implementation of ICustomUndoRedoFeature. Graphiti undo/redo
no longer needed
- Removed ability to disable Graphiti modification listener. No longer
necessary with new undo implementation.
- Added undo support for add/remove element
- Added undo support for ghosting
- Added undo support for setDiagramConfiguration
- Added ActionService, AgeAction, and ActionExecutor
- Implemented global undo/redo stacks.
- Reworked interface between AgeDiagram, GraphitiAgeDiagram, and
AgeDiagramEditor to use AgeAction interface.
- Undo/Redo actions for an editor are invalidated if the editor is
closed.
- Implemented switching to appropriate editor when undo/redo occurs.
- Implemented Undo/Redo of model changes.
- Updates DiagramExporterTest based on changes needed for the new action
system and to fail on exception.
- Added test for dirty state indicator.

### DIFF
--- a/ge/org.osate.ge/src/org/osate/ge/internal/selection/AgeBusinessObjectSelection.java
+++ b/ge/org.osate.ge/src/org/osate/ge/internal/selection/AgeBusinessObjectSelection.java
@@ -8,10 +8,18 @@ import java.util.function.Function;
 import java.util.stream.Stream;
 
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
 import org.osate.ge.BusinessObjectContext;
 import org.osate.ge.BusinessObjectSelection;
 import org.osate.ge.internal.services.AadlModificationService;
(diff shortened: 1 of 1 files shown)


ge/org.osate.ge.tests/src/org/osate/ge/tests/DirtyStateTest.java
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}, false, 0).click();
+	}
+
+	private void redo() {
+		bot.getActiveShell().menu().menu("Edit").menu(new CustomMatcher<MenuItem>("Redo Menu Item") {
+			@Override
+			public boolean matches(final Object item) {
+				final MenuItem menuItem = (MenuItem) item;
+				System.err.println(menuItem.getText());
```
RyanMcilnay: Remove print statement

philip-alldredge: Should be fixed.

PR #1351, opened 2018-06-20
Fix selection handling to use the current active selection.

### LINKED ISSUE #1313: Opening diagram from text editor
## Summary
Opening a diagram from the text editor always opens the package diagram the first time after opening the text editor.
It should open the associated diagram of the element that is selected in the text editor.

## Steps to Reproduce
1. Open text editor.
2. Right click element.
3. Click <em>Open Diagram</em>.

philip-alldredge:
Fix selection handling to use the current active selection. Closes #1313

### DIFF
--- a/ge/org.osate.ge/src/org/osate/ge/internal/ui/handlers/AgeHandlerUtil.java
+++ b/ge/org.osate.ge/src/org/osate/ge/internal/ui/handlers/AgeHandlerUtil.java
@@ -9,16 +9,18 @@ import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.ISources;
+import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.handlers.HandlerUtil;
 import org.osate.ge.internal.diagram.runtime.DiagramElement;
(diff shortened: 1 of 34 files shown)


ge/org.osate.ge/src/org/osate/ge/internal/ui/handlers/AgeHandlerUtil.java
```diff
@@ -9,16 +9,18 @@
...
 import org.osate.ge.internal.services.UiService;
 import org.osate.ge.internal.ui.util.SelectionUtil;
+import org.osate.ge.internal.ui.util.UiUtil;
 
 class AgeHandlerUtil {
 	// Returns the current selection as diagram elements.
 	// If one or more of the selected objects cannot be adapted to DiagramElement then an empty list is returned.
 	public static List<DiagramElement> getSelectedDiagramElements(final ExecutionEvent event) {
-		final ISelection selection = HandlerUtil.getCurrentSelection(event);
+		final ISelection selection = getCurrentSelection();
```
RyanMcilnay: This method is now the same as getDiagramElementsFromCurrentSelection() and does not use the event parameter.

philip-alldredge: @RyanMcilnay good catch. Should be fixed now.

PR #1358, opened 2018-06-21
Reworked DeleteHandler to group elements by resource.

### LINKED ISSUE #1357: Graphical editor deletes incorrect model elements.
In certain circumstances the graphical editor will delete the incorrect model element. This occurs when deleting multiple elements. If the model element's URI is based on its order in the model. This is common for many elements in the AADL model. 

The solution is to group model elements by resource when performing deletion of multiple elements and to perform a single modification for each resource. By retrieving the objects to be removed before modifying the model, the issue will be resolved. In cases where such behavior cannot be implemented, only deleting a single element should be allowed. This is already the case for annex elements but it will extend to objects which implement raw deletion operations such as packages.

philip-alldredge:
Reworked DeleteHandler to group elements by resource. Removes Graphiti delete feature in favor of allowing the delete handler to perform deletion logic. Simplifies case of deleting multiple elements. Closes #1357.


ge/org.osate.ge/src/org/osate/ge/internal/ui/handlers/DeleteHandler.java
```diff
@@ -81,57 +144,225 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
-					"Unable to retrieve pictogram element");
-			final DeleteContext deleteContext = new DeleteContext(pe);
-			final IDeleteFeature deleteFeature = fp.getDeleteFeature(deleteContext);
-
-			if (selectedDiagramElements.size() > 1) {
-				deleteContext
-				.setMultiDeleteInfo(new MultiDeleteInfo(false, false, selectedDiagramElements.size()));
+				deleteRaw(selectedDiagramElements.get(0), extensionService);
 			}
+			else if (anyIsInAnnex(selectedDiagramElements)) {
```
RyanMcilnay: else if is on the wrong line

philip-alldredge: @RyanMcilnay should be fixed now. I reran the formatter.

ge/org.osate.ge/src/org/osate/ge/internal/ui/handlers/DeleteHandler.java
```diff
@@ -81,57 +144,225 @@ public Object execute(final ExecutionEvent event) throws ExecutionException {
...
+		for (Object tmp = getContainer(bo); tmp != null; tmp = getContainer(tmp)) {
+			if (tmp instanceof AnnexLibrary || tmp instanceof AnnexSubclause) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	private static Object getContainer(Object bo) {
```
RyanMcilnay: Add final to bo

philip-alldredge: @RyanMcilnay fixed.

PR #1362, opened 2018-06-25
1327 emv2 unparse exceptions

### LINKED ISSUE #1327: Exception when serializing EMV2
Using the test case below, an exception is thrown. This was found while working with the graphical editor. In the graphical editor's case, the save appears to succeed but it is unclear what the side effects are. 

There are multiple problems demonstrated by the test case. The first error can be resolved by making EMV2AnnexUnparser public. Once that is done, other exceptions will be thrown. The latter exceptions only occur if there are more than one EMV2 subclause in the package.

Interestingly, the test does not fail even though many exceptions are thrown. I believe this is due to the exception being thrown in another thread. I am unaware of a mechanism to test for such errors.

## Steps to Reproduce
1. Run the test case below.

## Test Case
```
package org.osate.core.tests.issues

import com.google.inject.Inject
import com.itemis.xtext.testing.XtextTest
import org.eclipse.xtext.testing.InjectWith
import org.eclipse.xtext.testing.XtextRunner
import org.junit.Test
import org.junit.runner.RunWith
import org.osate.aadl2.AadlPackage
import org.osate.testsupport.TestHelper

import org.eclipse.xtext.resource.SaveOptions
import com.google.common.io.ByteStreams
import org.osate.testsupport.Aadl2UiInjectorProvider

@RunWith(XtextRunner)
@InjectWith(Aadl2UiInjectorProvider)
class Issue_NEW extends XtextTest {

	@Inject
	TestHelper<AadlPackage> testHelper;
	
	@Test
	def void issueNew() {
		val pkg = testHelper.parseString(aadlText)

		pkg.eResource().save(ByteStreams.nullOutputStream, SaveOptions.newBuilder().format().getOptions().toOptionsMap());
		 
	}
	
	val aadlText = '''
		package issue_new
		public
			system top
				annex EMV2 {**
				**};
			end top;
		
			system top2
				annex EMV2 {**
				**};
			end top2;
		end issue_new;
	'''
}

```

## Environment

* **OSATE Version**: Develop
* **Operating System**:  Windows 10
* **Java Version**: 1.8.0_141

lwrage:
Fixes #1327 
The NPEs in the formatter had the effect that formatting was skipped. The added test does the same without throwing an exception. It is not clear _why_ it happens that there is no region for an annex subclause.


lwrage:
build!

PR #1416, opened 2018-07-31
1415 test helper improvement

### LINKED ISSUE #1415: TestHelper does not handle multiple languages
Currently TestHelper parses Strings and files based on the extension of the primary language whose InjectorProvider is injected into the JUnit class. For parsing of files we want to be able to use it for other languages such as ReqSpec, Verify, etc.

reteprelief:
Fixes #1415


core/org.osate.testsupport/src/org/osate/testsupport/TestHelper.java
```diff
@@ -72,61 +75,65 @@ protected FluentIssueCollection testResource(Resource r) throws OperationCancele
...
-	 * @param path the file path starting with the test plugin name
-	 * @return the file content with unix line endings
+	 * load file as Xtext resource into resource set
+	 * @param filePath String
+	 * @param rs ResourceSet
+	 * @return
 	 */
-	public String readFile(String path) {
-		String result = "";
+	public Resource loadFileAsResource(String filePath, ResourceSet rs) {
```
lwrage: Please rename to loadFile.

PR #1423, opened 2018-08-01
1421 null global scope

### LINKED ISSUE #1421: ReqSpec does not resolve measurement units
Global scope does not get injected. It is used to build a list of all unit literals.

reteprelief:
Fixes #1421 

Added a ReqSpec test


core/org.osate.testsupport/src/org/osate/testsupport/TestHelper.java
```diff
@@ -72,61 +75,65 @@ protected FluentIssueCollection testResource(Resource r) throws OperationCancele
...
-	 * @param path the file path starting with the test plugin name
-	 * @return the file content with unix line endings
+	 * load file as Xtext resource into resource set
+	 * @param filePath String
+	 * @param rs ResourceSet
+	 * @return
 	 */
-	public String readFile(String path) {
-		String result = "";
+	public Resource loadFileAsResource(String filePath, ResourceSet rs) {
```
lwrage: Please rename to loadFile.

Issue #1438, opened 2018-08-09
Add documentation about working set support in the AADL Navigator.


lwrage:
child of #1014
child of #1371

PR #1439, opened 2018-08-09
Working set support for the AADL Navigator view

### LINKED ISSUE #1056: AADL Navigator Does not Properly Support Working Sets
- Selecting Top Level Elements->Working Sets with a working set selected results in only projects that aren't included in a working set appearing. 
    - Working sets do not appear because the org.eclipse.ui.navigator.resources.workingSets contentExtension isn't enabled.
- Unable to add Plugin_Contributions to working set. 

child of #1371

### LINKED ISSUE #1120: AADL Navigator Re-orders itself
## Summary
The AADL navigator's list of projects often re-sorts itself during tool execution. This is confusing and visually frustrating as I lose track of my projects. 

The ordering of projects in the AADL Navigator view should not change if the user is not changing the projects. 

## Steps to Reproduce
I installed OSATE 2.3.2 vfinal and the ALISA examples. Running the SCSCase causes the navigator list to re-order itself several times. 

Here's a video of the problem: https://youtu.be/tjE5eiGDBNY

## Environment

* **2.3.2 vFinal**:
* **Windows 7**:

lwrage:
fixes #1056 
fixes #1120 
The changes fix various issues with the old implementation of the AADL navigator where we didn't use the CNF correctly The working set support uses standard WS support from the Eclipse resource explorer.
We now show plugin contributions in each project. Adding a top node outside the working sets seems to be not worth the trouble.
Updated to the documentation are a separate issue (#1438).


sprocter:
Is this a big / important enough change to do an in-person review?

PR #1470, opened 2018-08-22
1465 exception check property constant

### LINKED ISSUE #1465: PropertyDoesNotApplyToHolderException in checkConstantProperty
When I have a property association for a feature I get the above exception in the CheckConstantProperty validator. It thinks the property is associated with a classifier. The property is only allowed on features.

lwrage:
fixes #1465


lwrage:
[build!]

PR #1598, opened 2018-10-16
Catch exception on findMarkers method on IResource. 

### LINKED ISSUE #1596: Marker based plugin analyses do not work when instance model and analysis results are not saved
For JUnit tests we were attempting to run without saving the instance model, the analysis results, and the assure instance. This causes problems for plugin analyses that still use markers to report issues as the IResource does not exist. Setting the save flag in tests does not help as the instance model is saved into the plugin, not into the workspace of the JUnit test.

reteprelief:
Fixes #1596 
In the long term we should replace the Marker based approach with AnalysisResult return values for all plugin analyses.


reteprelief:
Maven build failed in an area that I did not touch. Changes are adding a try catch in one method.

PR #1599, opened 2018-10-16
1584 compile without smaccm sources

### LINKED ISSUE #1584: Compile OSATE without smaccm sources
ALISA and EMV2 can execute Resolute. Compilation requires the sources from smaccm/smaccm in the workspace. 
Create a stub plugin for all classes and methods that are needed to compile without smaccm sources.
child of #920
requires #1106

lwrage:
fixes #1584 
requires #1594


lwrage:
I don't understand what should be changed...
The classes with comments are copied from the resolute plugin.

lwrage:
This change is just supposed to fix compilation errors. 1589 will still be applied to fix runtime issues.

PR #1628, opened 2018-11-07
Added table and graphical veiwer handler for faulttree files

### LINKED ISSUE #1627: Should be able to open existing fault tree in editor without having to regenerate it
Currently, the only way a fault tree can be viewed is by first regenerating it.
Add commands to open the fault tree in table format or in graphical format from the *.faulttree file.

reteprelief:
Fixes #1627


reteprelief:
I do not know the latest way we maintain menus in OSATE and how to do Open/Open with..
Can someone else take this over?

PR #1650, opened 2018-11-27
1649 bigdecimal in probability calculation

### LINKED ISSUE #1649: Probability computation for fault trees has rounding error due to use of Double
We should use BigDecimal to deal with rounding errors.
See #1643 as related issue.

reteprelief:
Fixes #1649


reteprelief:
There is still an issue with us implementing Real property values as double rather than BigDecimal. As a result the numbers users put into the model are not always precisely the ones retrieved for computation.

lwrage:
@reteprelief Can a user still open an existing fault tree if this change is merged?

reteprelief:
> 
> 
> @reteprelief Can a user still open an existing fault tree if this change is merged?

First, we regenerate the fault tree before showing it to the user.
Second, the representation for BigDecimal and Double are the same in the XML representation.

I did find an issue with generating the label for events in the graphical fault tree and pushed the correction out.

PR #1659, opened 2018-12-04
Contribute code generation property set

### LINKED ISSUE #1220: OSATE does not include Code Generation Annex property set
OSATE includes the ARINC653 property set from the ARINC 653 Annex, and the Data_Model and Base_Types files from the Data Modeling Annex. I would expect the Code_Generation_Properties property set from the Code Generation Annex to also be included, but it is not.

lwrage:
fixes #1220


lwrage:
@reteprelief Could you review, please?

PR #1674, opened 2019-01-16
1657 classifier view

### LINKED ISSUE #1657: AADL Classifier View
Need a view that shows all the members of a classifier (type/impl/feature group).  This is like a flattened view of the classifier, showing which ancestor a member is inherited from, whether it is refined, etc.  Also shows the complete ancestor hierarchy of the classifier.

EMF has the `EClass Information` view that does something similar.  Not sure if it is necessary to steal code from this view, but it's good for inspiration.

child of #1371

AaronGreenhouse:
Fixes #1657


AaronGreenhouse:
More correctly, the original test is broken.  I had to fix `FeatureGroupTypeImpl.getAllFeatures()` for #1657 and this revealed a bad assumption in the test.  The test is checking that an inverse feature group type has two and only two features.  I think the test is meant to verify that a refined feature from the feature group type being inverted is not included in the feature group.  Unfortunately, the old version of `getAllFeatures()` incorrecty ignored a feature from the root feature group type `BaseFG`.  My fixed version includes this feature, so the feature count is now 3.

AaronGreenhouse:
Fixed the test for issue 277.

PR #1677, opened 2019-01-30
Changed invoke methods in FlowLatencyAnalysisSwitch to return AnalysisResult

peterhfeiler:
Updated tests to not use LatencyAnalysisService.
Also updated GTSE not to use this class.


peterhfeiler:
Under wrong user ID

PR #1717, opened 2019-02-27
Fixes #1716 assure test problem

### LINKED ISSUE #1716: Assure test fails
Assure test fails due to alisa.contribution including code and property definitions that also exist in the test project. 
Remove them from the test project.

reteprelief:
Fixes #1716

Removed duplicate Physical and PhysicalResources from assure.tests. It is now provided through alisa.contribution.


lwrage:
The tests are not executed during the tycho build. Adding the plugin to the build lead to the build failing.

PR #1771, opened 2019-04-08
Allow diagram exporter to output to a stream.

### LINKED ISSUE #1835: Allow diagram exporter to output to a stream
## Summary

Currently, the diagram exporter utility can only write its output image to a Java File object. This makes it a pain to use with Eclipse IFile objects.

schwerdf:
Currently, the diagram exporter utility can only write its output image to a Java `File` object. This makes it a pain to use with Eclipse `IFile` objects.
This pull request adds two new methods to write the image to an `OutputStream` or an `ImageOutputStream` (the other two kinds of output supported by the backing image I/O library).
I added code to the diagram exporter's unit test to exercise the new methods, but was unable to get the test suite running (see recent posts on the mailing list), so due caution should be observed.

fixes #1835


philip-alldredge:
FYI. I plan to review this pull request this week.

ge/org.osate.ge/src/org/osate/ge/DiagramExporter.java
```diff
@@ -46,13 +48,8 @@
...
 public class DiagramExporter {
-	/**
-	 * Converts an AADL diagram file to a PNG image.
-	 *
-	 * @param  diagramFile the diagram file to be exported
-	 * @param  outputFile  the file the image will be written to
-	 */
-	public static void exportDiagramAsPng(final IFile diagramFile, final File outputFile) throws IOException {
+
+	public static org.eclipse.graphiti.mm.pictograms.Diagram readDiagram(final IFile diagramFile) throws IOException {
```
philip-alldredge: Is there a need for this to  be private? Since diagram exporter is part of the GE's API, I would prefer not to expose this method unless there is a compelling reason.

schwerdf: No, there is no compelling reason for that method to be public.

philip-alldredge:
@lwrage this is ready to be merged. I believed it when the build server was having issue.

PR #1791, opened 2019-05-03
Stand-Alone Annex Registry complete

### LINKED ISSUE #1387: Annex registry should have standalone mode
Currently the annex registry relies on the eclipse extension mechanism. to run in standalone mode there needs to be a registry that can be configured programmatically. This is similar to what EMF does for standalone operation.
child of #1538

AaronGreenhouse:
Fixes #1387


lwrage:
Where does the jar in jar loader zip file come from? It doesn't look like something that should be in the repository.

AaronGreenhouse:
> Where does the jar in jar loader zip file come from? It doesn't look like something that should be in the repository.

It comes from Eclipse, which plops it down when you generate a nested jar file.  I thought I got rid of all of those.  I just deleted the one still handing around.

AaronGreenhouse:
> The `jar-it.xml` files contains Windows specific and user specific paths such as `C:/Users/aarong/...` and `C:/SEI/Tools/...`

I've actually redone how this example works.  The ant script is gone now.

lwrage:
Merged manually

PR #1800, opened 2019-05-08
Fix build on JDK 11

### LINKED ISSUE #1794: Compilation errors on fresh development environment
- Created fresh development environment per instructions on osate.org.
- Received errors in org.osate.aadl.modelsupport and org.osate.analaysis.flow.tests projects.
- Those projects are configured to use Java 1.6 and 1.7 compliance levels but use functionality introduced in later versions of java.

## Environment
* **OSATE Version**: master

philip-alldredge:
Require JRE 1.8 for projects using 1.8 features. Fixes build on JDK 11.
Closes #1794.


lwrage:
@jjhugues Could you test with Java 12, please?

lwrage:
@philip-alldredge What happens if I use, for example, a collection method that was added after Java 8. Does compilation with Java 11 give me an error as the plugins are set to 1.8?

philip-alldredge:
@lwrage Yes, in that case it would provide an error just like it does now when using Java 8 constructs with the build path set to 1.6.  I assume this only effects newer JDKs because of some change in Eclipse or the JDK to enforce the specified build path. 

This change just adjusts these projects to be in line with other OSATE projects.

jjhugues:
@lwrage I just tried with Java 12, still got a couple of issues "release 6 is not found in the system"

I used the OSATE master branch with the previous commits.

lwrage:
@jjhugues I found that Java 12 support must be installed separately into eclipse 2019-03, it's not yet built in: https://marketplace.eclipse.org/content/java-12-support-eclipse-2019-03-411

PR #1814, opened 2019-05-19
Class cast exception due to missing end to end flow

### LINKED ISSUE #1812: ALISA: Verification fails with class cast exception when verification target does not exist in instance model
ALISA uses the enclosing component instance as target when the desired target is not found within the enclosing component instance.  That leads to class cast exception since the verification method expects an instance of the appropriate category.

This was triggered by the core instantiator not creating an end to end flow instance without informing the user. See #1809

reteprelief:
Fixes #1812


reteprelief:
I do not understand why the core.tests failed. They passed ok in the other three pull request.

Issue #891, opened 2017-08-27
Behavior annex errors are not propagated to the user


smithdtyler:
Errors reported by the behavior annex legality checker are generated but not reported to the user. For example, I (incorrectly) used a state transition with a `timeout` dispatch condition on a process behavior specification.

```
  process implementation processA.impl
    subcomponents
      tA: thread threadA;
    annex behavior_specification {**
    states
    s0: initial complete final state;
    s1: state;
    transitions
    s0 -[ on dispatch timeout ]-> s1;
    **};
  end processA.impl;
```

The OSATE workspace (I'm using the 2.2.3 testing build, but I've seen similar behavior on older versions) simply disables syntax highlighting for the annex code, reporting no useful information to the user. 

![screen shot 2017-08-27 at 2 18 33 pm](https://user-images.githubusercontent.com/8142684/29753245-b2d242f0-8b32-11e7-8d78-e603b244fde8.png)

Debugging from source indicates that the AadlBaLegalityrulesChecker determined my usage of the timeout keyword was incorrect. However, this error message does not appear in a problem marker or other user facing interface:

From AadlBaLegalityRulesChecker~560
```
        // Error case : it must only be declared for timed thread.
        this.reportLegalityError(tc, "The dispatch relative timeout and" +
              " catch statement must only be declared for timed thread: " +
              "Behavior Annex D.4.(L1) legality rule failed") ;
      }
```

smithdtyler:
Digging into this further, it appears that the behavior annex parser creates the appropriate error markers, but the markers are _sometimes_ deleted by OSATE. Placing a breakpoint at `org.osate.aadl2.modelsupport.errorreporting.MarkerParseErrorReporter:deleteMessagesImpl` indicates that when an AADL file is saved, markers are created and deleted multiple times, often asynchronously.

lwrage:
All this is a weird interaction between error reporting and AADL file expansion in the AADL navigator. It also leads to plenty of exceptions in the log file about missing markers, see #613.

Etienne13:
I reopen this issue since I could reproduce it in OSATE 2.5.0: with the example provided by Tyler, I see that OSATE error repporter is called from AadlBaLegalityRulesChecker~560. However the error is neither visible in the editor, nor in the problems view.

Etienne13:
[...]

PR #1875, opened 2019-07-02
Behavior annex plugin: improve syntax highlighting and error reporting

Etienne13:


Etienne13:
Behavior annex: 
 - enable syntax highlighting in case of syntax or name resolution errors
 - prevent the parser from reporting the entire annex block as erroneous on syntax errors

Etienne13:
@lwrage not sure why the last commit cannot be built. Is it because the static method convertToIResource (class OsateResourceUtil) is deprecated? Should I add an annotation or do you see a better way to do this?

Etienne13:
I found the recommendation to use toIFile instead of  convertToIResource in the comments. So I did the replacement.

Etienne13:
I reverted the last two commit because of the comments in issue #891

Etienne13:
@lwrage can you update me on the status of this pull request?

lwrage:
@Etienne13 We've merged #1905 to propagate annex errors and warnings. Could you update this PR and test that it still works, please?

Etienne13:
@lwrage Just did a merge of the master branch of osate2 in my fork. I had to fix a conflict and adapt my implementation. The PR is updated and tested: (i) syntax highlighting is visible in case of syntax/resolution error and (ii) the annexSection is set to null in case of error (thus removing the meaningless error message saying the initialState attribute cannot be null). However the checks performed by Jenkins failed; don't know why. Can you help me with that?

ba/org.osate.ba/src/org/osate/ba/texteditor/XtextAadlBaHighlighter.java
```diff
@@ -58,4 +60,24 @@ public void addToHighlighting(int annexOffset, int relativeOffset, int length, S
...
+
+  private static Map<AnnexSubclause, XtextAadlBaHighlighter> _highlighterPerAnnex = 
+      new HashMap<AnnexSubclause, XtextAadlBaHighlighter>();
+  
+  public static XtextAadlBaHighlighter getHighlighter(AnnexSubclause as)
+  {
+    if(_highlighterPerAnnex.get(as)==null)
+    {
+      XtextAadlBaHighlighter ht = new XtextAadlBaHighlighter();
+      _highlighterPerAnnex.put(as, ht);
```
lwrage: Why are highlighters stored in a has map? Each run of the parser creates new AnnexSubclause objects, so new highlighters will be created for each parser run.

lwrage:
There is a test failure. Project org.osate.ba.tests, file Issue1884Test.xtend. The test needs to be adapted to the changes in this PR.

ba/org.osate.ba/src/org/osate/ba/texteditor/XtextAadlBaHighlighter.java
Etienne13: The reason for using a map is the following: text to highlight in the BA subclause is identified during parsing (when the parser is executed from AnnexParserAgent class) but syntax highlighting is processed latter on (when highlightAnnexSubclause is called in class AnnexHighlighterProxy). So we need to save the locations to highlight during parsing and retreive them during highlighting.

The only shared (and stable) information between these methods invocation is the AnnexSubclause (or AnnexLibrary) object. This is why we store the locations to highlight in a map where the AnnexSubclause object is the key. This is also the reason why we store the current AnnexSubclause object in a global variable before starting the parsing. In AnnexParserAgent:

`if(defaultAnnexSection instanceof AnnexSubclause)
					AnnexUtil.setCurrentAnnexSubclause((AnnexSubclause) defaultAnnexSection);
`

I agree it is not very clean but given the methods provided by OSATE for syntax hihglighting of ANTLR based annexes I did not find a better solution so far.

ba/org.osate.ba/src/org/osate/ba/texteditor/XtextAadlBaHighlighter.java
```diff
@@ -58,4 +60,24 @@ public void addToHighlighting(int annexOffset, int relativeOffset, int length, S
               id));
   }
   
+  private XtextAadlBaHighlighter() {}
+
+  private static Map<AnnexSubclause, XtextAadlBaHighlighter> _highlighterPerAnnex = 
+      new HashMap<AnnexSubclause, XtextAadlBaHighlighter>();
```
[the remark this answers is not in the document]
Etienne13: I agree with the comment and the proposal, changed it in commit 2d1e528

Etienne13:
> There is a test failure. Project org.osate.ba.tests, file Issue1884Test.xtend. The test needs to be adapted to the changes in this PR.

I understand, but I do not know how to run this test. Can you explain me? Or do you want to update the test?

lwrage:
You can run all tests in a test plugin: right click on the project, select run as -> junit plug-in test
You can also run a single test when you right click on the xtend file. Tests are written in xtend but can also be written in plain Java (xtend is precompiled to Java).

Etienne13:
@lwrage thanks, the test is passing now.

PR #1879, opened 2019-07-08
1476 bend points bug

### LINKED ISSUE #1476: OSATE Graphical Editor should move bend points when both ends are moved together
When I select a group of objects in an AADL diagram and move them, anchor points on flows and connections between the objects don't move with them. By anchor points I'm referring to "turns" in the line connecting the two objects.

When _both_ objects connected by a flow/connection/etc are moved, all anchor points on that flow/connection should move too.

RyanMcilnay:
closes #1476


RyanMcilnay:
@philip-alldredge  ready for review again.

PR #1881, opened 2019-07-09
latency analysis on active etef only

### LINKED ISSUE #1872: Latency analysis is run on end to end flows not part of an SOM
We have a model with modes and some component or connection is not active in a mode.
An end to end flow can be defined without an in modes and it refers to components or connections that are not part of an SOM. In that case we should not run a latency analysis on it.

Example: session 13 solution of SEI course on AADL (\\ad.sei.cmu.edu\dfs\Groups\SSD\ACVI\sei-course\June 2019\Exercises)

reteprelief:
Fixes #1872


reteprelief:
good question about why the method exists. Some of its implementations have a comment that it was written when migrating from V1 to V2 to handle property retrieval.
I am making a pass at removing the method.

reteprelief:
I am also updating the forAllElement class to use isActive to filter on current SOM in the method processObject.

reteprelief:
Made the changes.

PR #1892, opened 2019-07-11
Updated create connection switch to used required_connection property 

### LINKED ISSUE #602: Required_Connection property not recognized
-- The latest version of OSATE issues an error message rather than
-- a warning for unconnected ports.  The Required_Connection property
-- should turn these error messages off, but it does not seem to.

package Unconnected
public
    data Message
    properties
        Data_Size => 32 bytes;
    end Message;

```
thread AThread
features
    msgIn: in data port Message;    
    msgOutA: out data port Message {Required_Connection => false;};
    msgOutB: out data port Message {Required_Connection => false;};     
end AThread;

process AProcess
features
    msgIn: in data port Message;    
    msgOutA: out data port Message {Required_Connection => false;};
    msgOutB: out data port Message {Required_Connection => false;};                     
end AProcess;

process implementation AProcess.Imp
subcomponents
    t: thread AThread;
connections
    cin: port msgIn -> t.msgIn;
    coutA: port t.msgOutA -> msgOutA;
    coutB: port t.msgOutB -> msgOutB;       
end AProcess.Imp;

system Unconnected
end Unconnected;

system implementation Unconnected.Imp
subcomponents
    p1: process AProcess.Imp;
    p2: process AProcess.Imp;
connections
    c12: port p1.msgOutA -> p2.msgIn;
    c21: port p2.msgOutA -> p1.msgIn;
end Unconnected.Imp;
```

end Unconnected;

AaronGreenhouse:
Fixes #602

### DIFF
--- a/core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateConnectionsSwitch.java
+++ b/core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateConnectionsSwitch.java
@@ -63,6 +63,7 @@ import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.osate.aadl2.AccessType;
+import org.osate.aadl2.BooleanLiteral;
 import org.osate.aadl2.BusAccess;
 import org.osate.aadl2.ComponentCategory;
 import org.osate.aadl2.ComponentImplementation;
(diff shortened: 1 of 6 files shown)


lwrage:
It doesn't look like this works with contained property associations. Also, what if the model is modal? Would it be better to add the check after instantiation?

AaronGreenhouse:
I was trying to work within the framework of how the existing error message was being generated.  I was aware modes might be a problem.  Forgot about contained properties.

The question I have overall, is this even the right way to check?   Right now it only complains if it follows a connection, and it cannot complete it.

But the spec implies that it should always be an a problem if there exists a feature that is not connected to something and the `required_connection` property is `true`.  So right now, if you never even try to build a connection at all, there would not be an error.

AaronGreenhouse:
See additional comments in the original issue #602.

lwrage:
Won't fix #602

PR #1898, opened 2019-07-12
1897 No latency report when no modes

### LINKED ISSUE #1897: Latency analysis does not produce report when model has no modes
OSATE 2.5.1-final staged
Use the session9 example for no report and session13 example for getting the report as the model has modes.

reteprelief:
Fixes #1897


reteprelief:
As you recommended we switched to using isActive to determine if an instance object is part of the current SOM. In the case of ETEFInstance isActive tests for som == null || inSystemOperationModes.contains(som).
It turns out inSystemOperationModes is empty when SOM is "no modes" instead of an SOM instance, while the SystemInstance object contains a SOM instance for "no modes" and is used to set current SOM.
This creates an inconsistency of how SOM(no modes) is handled.
The Latency handler counts on the doAaxlAction of one of the generic handlers, which sets the current SOM to SOM(nomodes).
The tests call an invoke method (same as ALISA verification calls). When I wrote that method I saw processObject in ForAllElement tests for som == null thus the invoke method leaves current SOM as null if the SOM is "nomodes".
The longer fix is to change the latency handler to also use invoke. It is more work as the handler assumes the internal latencyreport structure is returned and the spreadsheet is generated from that while the invoke method returns AnalysisResult.
I can go ahead and clean up the latency handler - if we do that it will only generate the csv file and not the xls file. I do have a method to generate csv from AnalysisResult, but no method to do the same for xsl. Is that ok? (WOuld allow us to get rid of the excel library that Julien had been using.

reteprelief:
I went ahead and changed the handler to use the same invocation methods as the tests.

PR #1903, opened 2019-07-16
Fix resource URI in SiriusUtil to load proper representation from aird file

GitNRG:
EcoreUtil appends suffix to resource URI that prevents loading proper resource later. This means that on model updates (say constants in aadl file) we'll keep showing original representation with old values and the only way to bypass this is to delete representations.aird file. With this update it will be enough to just instantiate model


reteprelief:
Cool.  I have been looking for the root cause of this issue for a while but not been able to put my finger on it. Thanks

PR #1914, opened 2019-07-23
1913 no undeveloped event

### LINKED ISSUE #1913: FTA includes undeveloped in propagation event when type is not propagated
An incoming error propagation may list several error types. Only a subset may be propagated from a particular source.
When FTA does a backward trace it not should include an undeveloped event for an error type that is included in the incoming propagation (willing to accept an error type) but not in the out going propagation (the type is not propagated). The undeveloped event intends to indicate that there is no connection (propagation path), which is not the case in this scenario.

reteprelief:
Fixes #1913


lwrage:
@reteprelief I merged in order, but there are still conflicts. Could you have a look and resolve the conflicts, please.

PR #1920, opened 2019-07-25
1918 alisa expressions

### LINKED ISSUE #1918: ALISA expression language cannot operations on range values
You can assign range values to variables, e.g., val myrange =  [20 ms .. 30 ms]

But you cannot operate on range values, e.g., 
value predicate max(#Compute_Execution_Time) <= 30 ms

It does not evaluate built-in functions min, max, abs, floor, ceil, round.

reteprelief:
Fixes #1918


reteprelief:
Fixed formatting.

PR #1930, opened 2019-08-03
1927 fta handles super types

### LINKED ISSUE #1927: FTA does not handle types with subtypes and type products well
When the fault tree starts out with or comes across an error type that has subtypes and during the trace it encounters one of the subtypes it does not continue the trace.
In the fault impact analysis we handle this scenario.
Similarly it handles type products by only propagating the first type in a type product. 
We want to propagate type tokens rather than error types.

reteprelief:
Fixes #1927

Modified a couple of tests to check on super types and on type products.


lwrage:
@jjhugues You know more about FTA than I so I've assigned the review to you.

PR #1938, opened 2019-08-08
1937 resolve subprogram access parameter

### LINKED ISSUE #730: Allow subprogram access parameter as flow start and end point
We need to include a subprogram access feature parameter as a flow source or destination point. See AADL V2.2

### LINKED ISSUE #1937: flow spec end to subprogram access parameter does not get resolved.
The standard allows a flow spec end to be a parameter in a subprogram access feature.

reteprelief:
Fixes #730


lwrage:
It looks like you directly changed the ecore file. That's a problem because it is generated. The correct way to change the metamodel is to update the UML model.

PR #1947, opened 2019-08-14
Abstract component as binding target

### LINKED ISSUE #1946: Connection binding property with abstract component as reference value gives erroneous error message
The reason is that abstract is not listed as category in acceptable reference values. 
In the applies to we do not have to explicitly list abstract while for specifying the reference type we do. This means, the target of binding properties can be abstract component without explicit specification.
[AbstractBinding.txt](https://github.com/osate/osate2/files/3501493/AbstractBinding.txt)
Allowing abstract is useful because we may have defined an abstract platform component.

The issue is relevant for connection, memory, and processor bindings. Function binding, which was introduced later, already has it correct.

reteprelief:
Fixes #1946

### DIFF
--- a/core/org.osate.workspace/resources/properties/Predeclared_Property_Sets/Deployment_Properties.aadl
+++ b/core/org.osate.workspace/resources/properties/Predeclared_Property_Sets/Deployment_Properties.aadl
@@ -12,7 +12,7 @@ property set Deployment_Properties is
 --- acceptable candidates.
 
 	Allowed_Processor_Binding_Class:
-		inherit list of classifier (processor, virtual processor, system, device)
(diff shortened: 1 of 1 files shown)


lwrage:
Do we need some tests that check if analyses work correctly if abstract is used?

reteprelief:
What I tried to address is that the standard is explicit about abstract always included in the applies to. However, it is silent as to whether the same holds for the reference (or classifier) property type. The easier solution is to explicitly add abstract in case of reference and classifier types. The alternative would be to implicitly include abstract which would require code changes.

lwrage:
This is also a language issue that should be looked at by the standard committee.

Issue #1124, opened 2018-04-10
Error on data access in the middle of end to end flow


joeseibel:
When an end to end flow refers to a data access, an error is reported complaining that data access references are only permitted at the beginning or end of an end to end flow. This validation doesn't make sense since it can't be found in the standard. The following model should be legal:
```
package connect_to_subcomponent
public
  system s1
  end s1;

  system implementation s1.i1
    subcomponents
      sub1: data;
      sub2: system s2.i1;
    connections
      conn1: data access sub2.f1 <-> sub1;
  end s1.i1;

  system s2
    features
      f1: requires data access;
  end s2;

  system implementation s2.i1
    subcomponents
      sub3: abstract a1;
      sub4: abstract a2;
    connections
      conn2: data access sub3.f2 -> f1;
      conn3: data access f1 -> sub4.f3;
    flows
      --Error: Illegal reference to 'f1'.  Cannot refer to a data access except for the first and last segment of an end-to-end flow.
      etef1: end to end flow sub3.fsource1 -> conn2 -> f1 -> conn3 -> sub4.fsink1;
  end s2.i1;

  abstract a1
    features
      f2: requires data access;
    flows
      fsource1: flow source f2;
  end a1;

  abstract a2
    features
      f3: requires data access;
    flows
      fsink1: flow sink f3;
  end a2;
end connect_to_subcomponent;
```

AaronGreenhouse:
This seems like a weird model.  Why are you communicating between two sibling components via the parent's feature?  How did this model come about?

Converting the model to use port features instead of shared access doesn't work either, instead we have the error "Couldn't resolve reference to EndToEndFlowElement 'f1'." on `f1`.

```
package connect_to_subcomponent2
public
  system xxx
  	features
  	  y: in out data port;
  end xxx;

  system s1
  end s1;

  system implementation s1.i1
    subcomponents
      sub1: system xxx;
      sub2: system s2.i1;
    connections
      conn1: port sub2.f1 <-> sub1.y;
  end s1.i1;

  system s2
    features
      f1: in out data port;
  end s2;

  system implementation s2.i1
    subcomponents
      sub3: abstract a1;
      sub4: abstract a2;
    connections
      conn2: port sub3.f2 -> f1;
      conn3: port f1 -> sub4.f3;
    flows
      --Error: Illegal reference to 'f1'.  Cannot refer to a data access except for the first and last segment of an end-to-end flow.
[...]

PR #1950, opened 2019-08-16
Removed overly restrictive validation check on data access in end to end flows

AaronGreenhouse:
Fixed Issue #1124


lwrage:
There are tests that expect the removed error message. They need to be adapted to the change.

PR #1955, opened 2019-08-19
1952 missing reference processor

### LINKED ISSUE #1952: MIPS budget analysis reports error if Reference_Processor is not specified
The online documentation indicates that the reference processor specification is not necessary if only one processor type is used. The reference processor is used in the conversion from execution time to MIPS. The MIPS budget analysis takes the execution time multiplied by period into account as actual value to compare to a specified MIPS budget value.

The tool checks for the presence of the Reference_Processor property and does not perform the analysis if not present.

reteprelief:
Fixes #1952 

This is one of the items reported by Steve and Alex probed about us fixing it.

I removed the validation for the presence of the Reference_Processor property. The property retrieval method is taking into account whether the property is present or not. 
I also removed some old code regarding MIPS conversion based on properties that no longer exist.
Conversion from execution time to MIPS is now based on the MIPS capacity and if present the calculated ratio between the reference processor capacity and target processor capacity. This is how the analysis is documented online. :-)


core/org.osate.xtext.aadl2.properties/src/org/osate/xtext/aadl2/properties/util/GetProperties.java
```diff
@@ -944,22 +930,6 @@ public static double getProcessorMIPS(final ComponentInstance processor) {
...
 	}
 
-	/**
-	 * compute MIPS for thread based on instructions per dispatch;
-	 *
-	 * @param threadinstance
-	 *            thread instance
-	 * @return MIPS or 0.0
-	 */
-	public static double getThreadExecutionIPDinMIPS(ComponentInstance threadinstance) {
```
lwrage: Public methods should not be removed without deprecating them fist. Also, add javadoc
`@deprecated will be removed in 2.6.0` (we search for such comments to find methods that can be removed). A comment why the deprecation and what to use instead shoud be there, too.

Same for other removed public methods

reteprelief:
Separating fix of issue from some dead code cleanup

PR #1981, opened 2019-09-02
Support probability precision and decimal format in FTA

GitNRG:
Let user choose probability format and specify desired result precision in Fault Tree Analysis


lwrage:
No response, closing.

PR #2004, opened 2019-10-01
2003 undersampling

### LINKED ISSUE #2003: Overestimation of flow latency computation with undersampling+sufficient output rate
## Summary

Flow latency computation does not differentiate cases associate with undersampling: when a component sends data with a higher frequency than the recipient AND the sender sends a data for each dispatch, then the period of the sender should be taken in the computation of the latency instead of the period of the recipient.

In case of undersampling, the flow latency analysis always takes the period of the recipient component. However, in some cases, it makes sense to take the period of the sender instead. The upper bound on the flow latency would be less pessimistic.

## Steps to Reproduce
1. Import the example below
2. Instantiate the system
3. Launch the flow latency analysis. Both flow (e2e_sample and e2e_change) have the same worst case latency but I believe the latency of e2e_sample should be 12ms lower than e2e_change.

The reason is that in e2e_sample, t1_sample sends a data after each dispatch (every 3ms) whereas in e2e_change, t1_change may or may not send a data after each dispatch. This is due to the property association "Output_Rate". Using this information (the sender sends at each dispatch at least as many value as needed by the recipient to start AND the frequency of the sender is higher than the frequency of the recipient), we obtain a less pessimistic upper bound for the end to end flow latency.

```aadl
package issue_fla_high_to_low_freq
public
	
	system s
	end s;
	
	system implementation s.impl
		subcomponents
			p: process p.impl;
	end s.impl;
	
	process p
	end p;
	
	process implementation p.impl
		subcomponents
			t1_sample: thread t_out {Period => 3 ms;};
			t1_change: thread t_out {Period => 3 ms;};
			t2: thread t_inout {Period => 15 ms; Deadline => 5ms;};
			t3: thread t_in {Period => 15ms;Compute_Execution_Time => 5ms..10ms;};
		connections
			c1: port t1_sample.p -> t2.p_i;
			c2: port t1_change.p -> t2.p_i;
			c3: port t2.p_o -> t3.p;
		flows
			e2e_sample: end to end flow t1_sample.s -> c1 -> t2.p -> c3 -> t3.s;
			e2e_change: end to end flow t1_change.s -> c2 -> t2.p -> c3 -> t3.s;
		properties
			Latency => 0 ms .. 100 ms applies to e2e_change, e2e_sample;
			Output_Rate=>[Value_Range => 1.0 .. 1.0; Rate_Unit => PerDispatch; Rate_Distribution => Fixed;] applies to t1_sample.p;
			Output_Rate=>[Value_Range => 0.0 .. 1.0; Rate_Unit => PerDispatch; Rate_Distribution => Fixed;] applies to t1_change.p;
	end p.impl;
	
	thread t_out
		features
			p: out data port;
		flows
			s: flow source p;
	end t_out;
	
	thread t_in
		features
			p: in data port;
		flows
			s: flow sink p;
	end t_in;
	
	thread t_inout
		features
			p_i: in data port;
			p_o: out data port;
		flows
			p: flow path p_i -> p_o;
	end t_inout;
	
end issue_fla_high_to_low_freq;
```

## Environment

* **OSATE Version**: 2.5.2
* **Operating System**: Linux

reteprelief:
Fixes #2003

### DIFF
--- a/analyses/org.osate.analysis.flows.tests/models/latencytest/issue2003.aadl
+++ b/analyses/org.osate.analysis.flows.tests/models/latencytest/issue2003.aadl
@@ -1,0 +1,51 @@
+package Issue2003
+public
+	
+	system s
+	end s;
+	
+	system implementation s.impl
+		subcomponents
+			p: process p.impl;
+	end s.impl;
+	
+	process p
+	end p;
+	
+	process implementation p.impl
+		subcomponents
+			t1_sample: thread t_out {Period => 3 ms;};
+			t2: thread t_inout {Period => 15 ms; Deadline => 5ms;};
+			t3: thread t_in {Period => 15ms;Compute_Execution_Time => 5ms..10ms;};
+		connections
+			c1: port t1_sample.p -> t2.p_i;
+			c3: port t2.p_o -> t3.p;
+		flows
+			e2e_sample: end to end flow t1_sample.s -> c1 -> t2.p -> c3 -> t3.s;
+		properties
+			Latency => 0 ms .. 100 ms applies to e2e_sample;
+	end p.impl;
+	
+	thread t_out
+		features
+			p: out data port;
+		flows
+			s: flow source p;
+	end t_out;
+	
+	thread t_in
+		features
+			p: in data port;
+		flows
+			s: flow sink p;
+	end t_in;
+	
+	thread t_inout
+		features
+			p_i: in data port;
+			p_o: out data port;
+		flows
+			p: flow path p_i -> p_o;
+	end t_inout;
+	
+end Issue2003;
--- a/analyses/org.osate.analysis.flows.tests/src/org/osate/analysis/flows/tests/Issue2003Test.xtend
+++ b/analyses/org.osate.analysis.flows.tests/src/org/osate/analysis/flows/tests/Issue2003Test.xtend
@@ -1,0 +1,59 @@
+package org.osate.analysis.flows.tests
+
+import com.google.inject.Inject
+import com.itemis.xtext.testing.XtextTest
+import org.eclipse.xtext.testing.InjectWith
+import org.eclipse.xtext.testing.XtextRunner
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.osate.aadl2.AadlPackage
+import org.osate.aadl2.SystemImplementation
+import org.osate.aadl2.instantiation.InstantiateModel
+import org.osate.analysis.flows.FlowLatencyAnalysisSwitch
+import org.osate.result.RealValue
+import org.osate.testsupport.Aadl2InjectorProvider
+import org.osate.testsupport.TestHelper
+
+import static org.junit.Assert.*
+
+import static extension org.junit.Assert.assertEquals
+
+@RunWith(typeof(XtextRunner))
+@InjectWith(typeof(Aadl2InjectorProvider))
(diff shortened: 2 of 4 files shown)


AaronGreenhouse:
Checks fail.

Also, I'm confused: The original issue makes the point the significant property association is `Output_Rate` on the data port.  I assume the `Value_Range` field, but as that is the only thing that is different between the `Output_Rate` values used on the ports in the original example.  I cannot find a description of exactly what that field is supposed to mean.  In any case, your fix looks solely at the `Period` property association of the components.  

Also, I'm not sure when this problem started, but when I try to run the latency analysis, I get an exception that it doesn't have permission to write the output file.  I know I was able to do this a few months ago.

```
!ENTRY org.eclipse.egit.core 1 0 2019-10-02 13:40:48.850
!MESSAGE Using Apache MINA sshd as ssh client.
java.lang.IllegalStateException: Cannot modify resource set without a write transaction
	at org.eclipse.emf.transaction.impl.TransactionChangeRecorder.assertWriting(TransactionChangeRecorder.java:349)
	at org.eclipse.emf.transaction.impl.TransactionChangeRecorder.appendNotification(TransactionChangeRecorder.java:303)
	at org.eclipse.emf.transaction.impl.TransactionChangeRecorder.processResourceNotification(TransactionChangeRecorder.java:273)
	at org.eclipse.emf.transaction.impl.TransactionChangeRecorder.notifyChanged(TransactionChangeRecorder.java:239)
	at org.eclipse.emf.common.notify.impl.BasicNotifierImpl.eNotify(BasicNotifierImpl.java:424)
	at org.eclipse.emf.common.notify.impl.NotifyingListImpl.dispatchNotification(NotifyingListImpl.java:261)
	at org.eclipse.emf.common.notify.impl.NotifyingListImpl.addUnique(NotifyingListImpl.java:294)
	at org.eclipse.emf.common.util.AbstractEList.add(AbstractEList.java:304)
	at org.osate.analysis.flows.FlowLatencyUtil.saveAnalysisResult(FlowLatencyUtil.java:544)
	at org.osate.analysis.flows.handlers.CheckFlowLatency.finalizeAnalysis(CheckFlowLatency.java:130)
	at org.osate.ui.handlers.AbstractInstanceOrDeclarativeModelReadOnlyHandler.doAaxlAction(AbstractInstanceOrDeclarativeModelReadOnlyHandler.java:125)
	at org.osate.ui.handlers.AbstractAaxlHandler.processAaxlAction(AbstractAaxlHandler.java:482)
	at org.osate.ui.handlers.AbstractAaxlHandler.actionBody(AbstractAaxlHandler.java:191)
	at org.osate.ui.handlers.AaxlReadOnlyHandlerAsJob$ActionAsJob.runInWorkspace(AaxlReadOnlyHandlerAsJob.java:124)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:42)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

reteprelief:
Aaron,
OutputRate is not interpreted by the latency analysis. 
Etienne used output rate in a funny way  0..1 to mean either no value or 1 value. He also interpreted it to mean every other value is no value. 
This is better specified by a different period of the source thread.
I'll check about the exception.

reteprelief:
test is rerunning. I had the aadl file name spelled as Issue2003.aadl and not issue2003.aadl
For some reason that did not cause a problem on Windows.

PR #2024, opened 2019-10-15
Some fixes for alisa arithmetics

### LINKED ISSUE #2031: Arithmetic bugs in ALISA
When dividing 2 values with same dimension: the result should have no unit.
Multiplication of integer values is broken: a * b is a^2.
Absolute value function is broken: abs(x) is -x.

paolo-crisafulli:
Hi,
This branch contains 3 fixes for arithmetics in reqspec files:
- Fix the resulting unit when dividing 2 values with same dimension: the result should be adimensional.
- Fix multiplication of integer values.
- Fix the absolute value function.

fixes #2031


lwrage:
@paolo-crisafulli  Thank you for this PR.

Issue #2056, opened 2019-11-09
Add RAMSES to the list of installable components


lwrage:
## Summary

Allow installation of the RAMSES code generator via the additional OSATE component installation dialog.

## Environment

* **OSATE Version**: 2.6.0
* **Operating System**: all

lwrage:
Waiting for @Etienne13 to provide the required information.

lwrage:
@Etienne13 Has there been any progress?

PR #2069, opened 2019-11-21
Fix issue #2056

Etienne13:


Etienne13:
I don't understand why the build of this PR fails... @lwrage can you tell me more?

PR #2135, opened 2019-12-20
2123 skipped steps in trace

### LINKED ISSUE #2123: Fault trace generation does not show all steps in trace
In some cases the trace uses a substep as step if there is only one substep.
Those situations will be removed during optimization.
Example: 
The example submitted by Brian Larson in issue #2111 
The trace for undetected failure in iPCA_Safety.i only shows the error event and the composite error state, not the state of the subcomponent that leads to the error event.

reteprelief:
Fixes #2123


lwrage:
@reteprelief FTA Tests fail. Could you check, please?

reteprelief:
Fixed the issue but cannot push the fix. I did a pull on the master. It has error in the ReqTrace manifest.
When I merged the master into 2123 I could not push it out - something seems to be out of sync (it shows 5 changed with down arrow and 13 with up arrow).

reteprelief:
Reset my workspace and was able to make the mod and push it out.

Issue #2146, opened 2020-01-03
Unable to reference port in feature group from Behavior Annex specification.


smithdtyler:
## Summary

In 2.6.0 I was able to reference ports in feature groups as part of transition guards in the AADL behavior annex. That ability seems to have been lost in 2.6.1. @Etienne13 indicated this is likely a regression. 

See https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/osate/zeZozfceOqI

## Steps to Reproduce
See AADL model below.

```aadl


feature group RodentsRequired

features 

Hamster: in event data port;

    end RodentsRequired;

    

    feature group RodentsProvided

features 

Gerbil: out event data port;

    end RodentsProvided;

  

 thread group threadA

    features

      inputA: in feature group RodentsRequired;

      outputA: out feature group RodentsProvided;

  end threadA;

  

  thread group implementation threadA.impl

    subcomponents

      t0 : thread;

    annex behavior_specification {**

    states

    s0: initial state;

    s1: final state;

    transitions

    s0 -[inputA.Hamster=1]-> s1 {outputA.Gerbil:=1};

    **};

  end threadA.impl;


```
_Error message: Wrong type in value variable stop rule; expected types are: data subcomponent or data access or behavior variable or data access feature prototype or in parameter or in port or in port prototype or iterative variable_

## Environment

* **OSATE Version**: 2.6.1
* **Operating System**: Windows 10

lwrage:
@Etienne13 Do you think you can fix this for the next release (end of Jan 2020)?

Etienne13:
This issue should be fixed after integrating PR #2152.

Issue #2147, opened 2020-01-03
Unable to use on port as transition condition in the Behavior Annex


smithdtyler:
Via @Etienne13 

> Input ports may be used in transition of such components, in particular in order to represent mode transitions. In this case, the syntax allows to define the transition condition as follows:
> 
> s0 -[on input]-> s1;

> It seem the current version of the BA plugin does not support this syntax. You can fill in an issue on github and I will have a look at the problem.

## Steps to Reproduce
See example model below

```aadl
 thread group threadA

    features

      input: in event data port message_pm;

  end threadA;


  thread group implementation threadA.impl

    subcomponents

      t0 : thread;

  

    annex behavior_specification {**

    states

    s0: initial final state;

    s1: state;

    transitions

    s0 -[ on input]-> s1;

    **};

  end threadA.impl;
```

## Environment

* **OSATE Version**: 2.6.1
* **Operating System**: Windows 10

Etienne13:
This issue should be fixed after integrating PR #2152.

Etienne13:
@smithdtyler to be consistent with the BA standard:
- s0 should be the identifier of a mode in threadA.impl
- s0 should be a complete state

here is the result:

```aadl
  thread group implementation threadA.impl
    subcomponents
      t0 : thread;
    modes
      s0: initial mode;
    annex behavior_specification {**
    states
      s0: initial complete final state;
      s1: state;
    transitions
      s0 -[ on input]-> s1;
    **};
  end threadA.impl;
```

PR #2152, opened 2020-01-07
Support for mode switch transitions in BA

Etienne13:


lwrage:
Can you add a JUnit test for #2147, please?

Etienne13:
It is present in the first commit

lwrage:
I have no idea how I missed that,,,

lwrage:
@jjhugues Could you review this, please. It should fix #2146 and #2147.

jjhugues:
@Etienne13 , this patch addresses #2146 and #2147 but I miss a test for mode transitions.
I would have expected a test that uses AADL core mode, is that correct?

Etienne13:
@jjhugues , this is a good question, and I was not completely sure on how to proceed. In BA, transitions triggered by in event [data] ports in "non-dispatchable" components aim at refining mode transitions. However, I was not sure whether I should enforce this in the BA plugin or not. By enforcing it, the plugin becomes more rigorous and less flexible. Enforcing the constraint would be consistent with the way the plugin is implemented for the other parts of the standard.

What do you think?

jjhugues:
@Etienne13 , I was pointing at a more simple issue: I see no test with modes that would test the code you added for modes. Did I miss it? You added two tests for other issues that this PR closes

lwrage:
@Etienne13 If this PR does more than fix 2146 and 2147, could you create an issue for that, please?

Etienne13:
@jjhugues I did not add code for modes per se, but I added code for supporting transitions in BA which refine mode transitions from the core standard. This is the fix for #2147, where in event [data] ports are used in transition conditions. The only situation I know of that would allow this is when the source state is a mode (the state has the same identifier as a mode of the enclosing component).

To be more consistent, I added a mode in the test case for issue #2147. This way the AADL model actually covers the type of models covered by the fix.

@lwrage ok, I will open an issue to enforce the input state of such transitions (see comments above) has the same identifier as a mode of the enclosing component.

jjhugues:
@Etienne13 , can you either update this PR or close it and do a new one? The test for Issue 2417 is failing

PR #2157, opened 2020-01-10
Report a warning on dispatch conditions in abstract components

### LINKED ISSUE #2154: Abstract component and on dispatch
The BA parser has now and extension of D.3 (L5) that forbids abstract components to have on dispatch.

abstract components cannot contain a dispatch condition in any of its transitions: they cannot be dispatched (extention of Behavior Annex D.3.(L5) legality rule).	

This seems too strong, abstract components exist for the purpose of specification, they will be later refined. 

This rules forbids the construction of abstract functions that would later be refined as an entity that can be dispatched. Can you turn this into a warning or info ?

Etienne13:
Report a warning when dispatch conditions are used in abstract components, and an error for other component categories on which the dispatch_protocol property cannot be applied.

Fixes #2154.


lwrage:
@Etienne13 The Dispatch_Protocol property _can_ be applied to abstract components. In fact, _all_ properties apply to abstract components (see 11.1.2 L1 in the standard).

jjhugues:
@lwrage before reviewing the patch, can you investigate the build error? Looks like a setup issue

lwrage:
It is a network configuration issue. Waiting for IT to fix it. You'll need to run a local build or wait until sometime next week.

lwrage:
@jjhugues I've implemented a workaround. The build no longer needs access to remote p2 repos.

PR #2179, opened 2020-01-24
Fixed data access connections to subprograms

### LINKED ISSUE #2161: Some data-access connections are not being instantiated.
## Summary
When instantiating the AADL model below using the latest development version of OSATE, a data-access connection within the model is not being instantiated.

When instantiating this model with OSATE 2.6.1, the instance file contained a connection instance corresponding to the data-access connection `s1`:

```
    <connectionInstance name="shared -> t1.shared" complete="true" kind="accessConnection" destination="//@componentInstance.0/@componentInstance.1/@featureInstance.0" source="//@componentInstance.0/@componentInstance.0">
      <connectionReference context="//@componentInstance.0" source="//@componentInstance.0/@componentInstance.0" destination="//@componentInstance.0/@componentInstance.1/@featureInstance.0">
        <connection xsi:type="aadl2:AccessConnection" href="../DataAccessConnectionTest.aadl#/0/@ownedPublicSection/@ownedClassifier.7/@ownedAccessConnection.0"/>
      </connectionReference>
    </connectionInstance>
```

## Steps to Reproduce
1. With the AADL model below, instantiate the system implementation `DACT.impl`.
1. Open the instance file.
1. Note that there is no connection instance of any kind in the file.

```aadl
package DataAccessConnectionTest
public
	data Shared		
	end Shared;
	
	data implementation Shared.impl
	end Shared.impl;
	
	subprogram RequiresData
	features
		shared: requires data access Shared;
	end RequiresData;
	
	subprogram implementation RequiresData.impl
	end RequiresData.impl;
	
	thread ThreadRequiresData
	features
		shared: requires data access Shared;			
	end ThreadRequiresData;
	
	thread implementation ThreadRequiresData.impl
	subcomponents
		sp: subprogram RequiresData.impl;
	calls main: {
		s1 : subprogram sp;
	};		
	connections
		dc: data access shared -> sp.shared;
	end ThreadRequiresData.impl;		
	
	process Proc	
	features
		externalData: requires data access;					
	end Proc;	
	
	process implementation Proc.impl
	subcomponents
		shared: data Shared.impl;
		t1: thread ThreadRequiresData.impl;
	connections
		s1: data access shared -> t1.shared;
	end Proc.impl;	
		
	processor P
	end P;
	
	processor implementation P.impl
	end P.impl;

	system DACT
	end DACT;

	system implementation DACT.impl
	subcomponents
		proc: processor P.impl;
		app: process Proc.impl;
	properties
		Actual_Processor_Binding => (reference (proc)) applies to app;		
	end DACT.impl;		

end DataAccessConnectionTest;
```

## Environment

* **OSATE Version**: 2.7.0-v20201007-1612
* **Operating System**: Windows 7, macOS 10.14

AaronGreenhouse:
Fixes #2161


stevevestal:
A couple of general comments that might be relevant to long-term plans for instantiating end-to-end connections.

The earlier convention was to have instantiated connections go from thread port to thread port, but not dive down into thread internals.  Tool developers had to do their own stuff to get internal details as needed (which we are revisiting in the case of access connections to data sub-components).  Many kinds of connections can do this, not just data access connections.  Going forward, if the plan is to move in general towards instantiating connections from true first-feature to last-feature, something to consider is the impact on legacy tools from a variety of third-party developers.

In such cases, it is nice to be able to easily determine the containing thread and the feature of the containing thread through which a connection passes (e.g., useful to obtain properties such as processor bindings that do not apply to data features and thus are not inheritable but might be relevant to some analyses); and to easily determine the containing process and the feature of the containing process through which a connection passes (e.g., the AADL ARINC 653 Annex has places where it says, or at least strongly implies, that properties of the process rather than the thread ports specify certain things about connections between A653 partitions).

lwrage:
@stevevestal Currently we have no plans to change connection instantiation further. Regarding the impact on existing tools: If our change breaks your tools, let's discuss via the google group.

PR #2189, opened 2020-02-05
Created new Instantiate and Reinstantiate commands and handlers

### LINKED ISSUE #1553: Redesign reinstantiation handlers
The current state of instantiation and reinstantation handlers is confusing and it seems that it was created in an ad-hoc manner. We should think about what would make sense.

There are three handlers for instantiation: `InstantiateHandler`, `ReinstantiateAadl`, and `ReinstantiateHandler`. We should use more meaningful names so that it can be obvious what the difference between `ReinstantiateAadl` and `ReinstantiateHandler` is. Some of the functionality is confusing as well.

- `InstantiateHandler`: This is in the outline view's context menu with the label "Instantiate" and is invoked on one `ComponentImplementation`.
- `ReinstantiateAadl`: This is in the main menu with the label "Reinstantiate All Instance Models". The behavior is dependent on the selection. If the selection contains instance `IFile`s, then only those instance files are reinstantiated. This is confusing because the behavior does not match the label and it will not reinstantiate all instance models in the workspace. If the selection does not contain any instance files, then all instances models in the workspace will be reinstantiate. This handler needs to change because the behavior is confusing and not obvious to the user. "Reinstantiate All" should not care about the selection.
- `ReinstantiateHandler`: This is in the navigator's context menu with the label "Reinstantiate" and is invoked on one `IFile` which is an instance file. This could be modified to allow multiple instance files to be selected.

One functionality to think about is reinstantiating all instances in a selected `IContainer` or working set. There could be a dialog showing which files were reinstantiated or indicating that there were no instance files contained in the selection.

We are also not consistent with how progress is handled. `InstantiateHandler` is run in a `ProgressMonitorDialog`. `ReinstantiateAadl` is run in a `WorkspaceJob`. `ReinstantiateHandler` is executed directly without any progress reporting.

AaronGreenhouse:
Fixes #1553


AaronGreenhouse:
Need to update the user guide!

AaronGreenhouse:
Updated user guide.

AaronGreenhouse:
> I found two situations where reinstantiate isn't enabled:
> 
> * when the selection in the navigator contains a closed project, e.g., one open and one closed project (the command should be enabled and closed projects should be ignored)
> * when I select a working set (the command should process all projects in the working set, ignoring closed projects in the working set)

Fixed this.

Issue #1543, opened 2018-09-21
Add action to add all elements that are connected to a selected element


lwrage:

RyanMcilnay:
@philip-alldredge @lwrage Current thoughts for implementation is to add a context menu item for ConnectionEnds/ConnectionInstanceEnds that will show continuous connected elements in the ancestors and children.

@lwrage Any thoughts or concerns?

lwrage:
I'm not sure I understand what you mean by ancestors and children.

philip-alldredge:
@lwrage  @RyanMcilnay can clarify but...

We are ultimately trying to decide how this feature should work given the name of the issue. The current idea is that if a connection end is selected and the menu item is used, it will show all connections that reference it and the other end of the connection. 

The question we had was whether selecting a subcomponent and using the menu item should show features and everything connected to those features. My feeling is that behavior would be too broad and that it shouldn't do that. However, we don't know the original source of this request. We can also add that capability later once users have experience with this feature.

lwrage:
@philip-alldredge IIRC, what you just described as too broad was the original idea behind the issue. This looks like a topic for a telecon.

philip-alldredge:
If that's the case then that is fine. The too broad comment was mostly just my initial feelings on it. I didn't want to introduce features in the absence of a well defined scope.  

@lwrage @RyanMcilnay I'd recommend setting up a telecon between the two of you. I will attend as my availability allows but I would prefer not to hold things up on my account. I'd like to update this issue when it is resolved though. 

Is the intention for this to only be applicable to subcomponents or to both subcompoennts and any connected elements?

RyanMcilnay:
@lwrage Disregard the child/ancestor comment, it was based off of a misunderstanding.

I put a few examples together of how this action would behave.  The action would be applicable to Subcomponents and ConnectionEnds.

If this is an acceptable approach, I'll continue in this direction.
If there are any concerns, just send me and email we can discuss on the phone.

This is the full diagram.
![diagram](https://user-images.githubusercontent.com/8505197/73975832-bab30e80-48ec-11ea-9d05-f8ddb701b387.PNG)

Before showing connections of ss1.
![hidden](https://user-images.githubusercontent.com/8505197/73975761-9d7e4000-48ec-11ea-9909-9d37ad1eff01.PNG)

After showing connections of ss1.
[...]

PR #2210, opened 2020-02-25
1543 show connected elements

RyanMcilnay:
#1543 Show Connected Elements


philip-alldredge:
- The menu option isn't available for subcomponents that don't have a component implementation. The menu option should be available for all subcomponents which have a Classifier as well as any subcomponents which may be a connection end(data and bus?).
- Multiple things are being added to the undo stack. When the operation is performed the undo menu should have a single action added to it and it should be labeled appropriately. "Show Connected Elements".
- Layout isn't being triggered properly. Partial layout. Fixing the undo stack issue may fix this one as well.
- Don't reference DiagramConfigurationDialog.Result. It's not needed and doesn't make sense to reference part of that diagram.
- Fix spelling for enableComponentImplemenationConnections.

I'll do a more full review once things are resolved.

RyanMcilnay:
@philip-alldredge Ready for review

Issue #1544, opened 2018-09-21
Add action to show all elements that participate in a flow or mode to a diagram


lwrage:

philip-alldredge:
@lwrage is this idea for this to be just a shortcut for manually adding elements to the diagram or something that is more dynamic/persistent? In other words, if the flow or mode changes, should the set of elements shown in the diagram change as well or should the user have to manually add the new elements/repeat the operation?

RyanMcilnay:
@lwrage Current thoughts for implementation are to add a toolbar button that will show the elements that participate in the currently selected flow in the highlight flow combobox.

Modes are selectable on the diagram so a context menu option would be added to show in mode elements.

Does this seem like an acceptable approach?

Notes: Disable/Enable button when whole/partial flows are selected

philip-alldredge:
That seems like it would acceptable to me.

philip-alldredge:
@RyanMcilnay I noticed that the flow drop down does not show flows when there are no elements in the flow currently visible. For example if a subcomponent contains an end-to-end flow but its contents are not being displaced, the flow is not selectable. We need to adjust this to show such flows. I believe a general rule would be that if the container is visible, then we should show it.

philip-alldredge:
@RyanMcilnay i'm working with an out of date version, but please verify that undo and redo works correctly. It should be labeled based on the operation being performed and there should only be one undo required to undo the entire operation.  This is not the case with the currently pushed version.

lwrage:
Looks good to me.

PR #2211, opened 2020-02-25
1544 show flow or mode elements

RyanMcilnay:
#1544 Show flow or in mode elements


lwrage:
Something is missing, there should be changed plugin versions.

philip-alldredge:
- Multiple actions are being to the undo stack. Same issue as in pull request for #1543
- Layout not being executed in some cases? Related to undo stack?
- "Show In Mode Elements". Should be renamed. "Show Elements In Mode". Also update documentation. Suggest reword. "will show the elements within the mode's container which are active in the selected mode".
- Revert changes to the DiagramConfiguraitonDialog. Do not reference it.

I'll do a more full review once the issues are resolved.

PR #2214, opened 2020-02-27
2199 ba struct union element containment

### LINKED ISSUE #2199: Behavior Annex: Null pointer exception when calling getContainingFeature() on an instance of  StructUnionElement
## Summary
We use the EFM Compare framework to compare two AADL models having BA annex clauses. A null pointer exception occurs when comparing the models as shown in the trace below:

	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eContainingFeature(BasicEObjectImpl.java:908)
	at org.eclipse.emf.compare.match.eobject.EUriFragmentFunction.apply(EUriFragmentFunction.java:49)
	at org.eclipse.emf.compare.match.eobject.EUriFragmentFunction.apply(EUriFragmentFunction.java:1)
	at org.eclipse.emf.compare.match.eobject.URIDistance.retrieveFragment(URIDistance.java:187)
	at org.eclipse.emf.compare.match.eobject.EditionDistance$1$1.matchingURIs(EditionDistance.java:143)
	at org.eclipse.emf.compare.utils.EqualityHelper.matchingEObjects(EqualityHelper.java:251)
	at org.eclipse.emf.compare.utils.EqualityHelper.matchingValues(EqualityHelper.java:170)
	at 
...

This occurs when calling eContainningFeature() on an instance of the StructUnionElement class of the BA metamodel. 

The problem is due to the instance being added to the StructUnionElementHolder parent using the eBasicSetContainer() method passing the AadlBaPackage.STRUCT_UNION_ELEMENT_HOLDER as containing feature ID (see AadlBaTypeChecker class line 1357). 

This is wrong since the passed ID is the ID of the StructUnionElementHolder class in the Ecore meta-model. As a matter of fact, there is a problem in the BA metamodel since there is no feature in the StructUnionElementHolder  class to contain the StructUnionElement. The "element" reference of the StructUnionElementHolder class cannot be used for that since it is not containement.

I suggest adding an explicit reference in the StructUnionElementHolder class to contain StructUnionElement elements and modify the AadlBaTypeChecker class to use the proper reference constant.

## Environment

* **OSATE Version**: 2.6.0
* **Operating System**:

Etienne13:
fixes #2199


lwrage:
@Etienne13 Who can review this?

Etienne13:
@lwrage I don't really see who can do this specifically ... Which qualifies anyone who would be willing to do the review ;). The only required skill/knowledg is EMF.

lwrage:
@Etienne13 Are you going to add a JUnit test case, and did @dblouin test it?

dblouin:
Hi all,

We have tested the fix and it solves the problem. Thanks.

PR #2215, opened 2020-02-28
Changed PropertiesProposalProvider.showCrossReference() to use better type equality

### LINKED ISSUE #2073: Content assist should propose property constants
## Summary

It would be convenient to have property constants that have the correct type proposed in content assist. Currently property constants are not proposed on Ctrl+Space.

## Environment

* **OSATE Version**: 2.6.0
* **Operating System**: all

AaronGreenhouse:
Fixes #2073


AaronGreenhouse:
> I found a case that doesn't seem to work: property type R is a record, C is a constant of type R and property L is a list of R. C is not proposed as a list element when writing a property association for L.
> I didn't try this scenario with a list of a base type, so that may or may not work.

Yes, I forgot about the case of proposing values inside a list.

AaronGreenhouse:
> I found a case that doesn't seem to work: property type R is a record, C is a constant of type R and property L is a list of R. C is not proposed as a list element when writing a property association for L.
> I didn't try this scenario with a list of a base type, so that may or may not work.

Fixed this.

lwrage:
@AaronGreenhouse please fix the build, https://github.com/osate/osate2/wiki/Managing-Feature-and-Plugin-Versions#troubleshooting should help

AaronGreenhouse:
> @AaronGreenhouse please fix the build, https://github.com/osate/osate2/wiki/Managing-Feature-and-Plugin-Versions#troubleshooting should help

Fixed this

Issue #1202, opened 2018-04-25
Support Creating a Flow Implementation without Flow Specification being Visible


philip-alldredge:
<a href="https://github.com/philip-alldredge"><img src="https://avatars2.githubusercontent.com/u/1376364?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [philip-alldredge](https://github.com/philip-alldredge)**
_Wednesday Feb 14, 2018 at 16:35 GMT_
_Originally opened as https://github.com/osate/osate-ge/issues/294_

----

Support creating a flow implementation without flow specification being visible. The current flow tools requires selecting the flow spec graphically.

philip-alldredge:
Allow flow specification and other elements to be selected from the outline with link with editor off and when the model elements are not in the diagram. Update prompts and/or user guide to make it clear to the user this is allowed.

PR #2228, opened 2020-03-06
1774 ge flow creation fix

### LINKED ISSUE #1774: End to End flow creation in Graphical Editor requires extra clicks
I am using OSATE Stable 2.4.0 on Windows 7. 

I have a graphical AADL model with flow sources, sinks, connections, and paths. Per the instructions in the Help menu, I selected a _flow source_ and clicked on the "Create End to End  Flow Specification" button, which launches the dialog shown below:

Screen Shot 2019-04-10 at 12 33 58 PM

When this dialog appears, the flow source I originally selected is un-selected. I provide a name for the end to end flow and click on the flow source again, but it is not added to the flow specification. I have to select some _other_ element in the model first, then re-select the flow source in order for it to be added to the flow specification. 

The end to end flow dialog should automatically state with the flow source that was selected when the user clicked the "Create end to end flow specification" button.

RyanMcilnay:
#1774 #1202 

Initial selection from tool dialogs fixed and elements can be selected using the outline view (elements that are not on the diagram).


RyanMcilnay:
@philip-alldredge Good catch, should be fixed now.

Also, I fixed the error message when undo is performed. Clearing it because selection is cleared.  This only affected the EndToEnd tool.

Issue #2083, opened 2019-11-27
Add dialog proposing to save files prior to instantiation


jjhugues:
Currently, one can instantiate a system with unsaved files. This can create inconsistent results from the user perspective.

A check could be added to look for open files, and a pop-up be raised to propose saving all files prior to process.

AaronGreenhouse:
UI Spy: **ALT+SHIFT+F1**

AaronGreenhouse:
Two main problems for this:
1. How to find out what files are open and need to be saved.
2. How to use the same dialog box that Eclipse already uses (for example when you have open files and select a launch configuration.)

AaronGreenhouse:
According to the UI Spy the dialog in question is `org.eclipse.debug.internal.ui.launchConfigurations.SaveScopeResourcesHandler.ScopedResourcesSelectionDialog`.

Seems to be part of a large operation.

Technically we should rely on this because it's in an "internal" package?

AaronGreenhouse:
This method or a clone of it is handy (from `SaveScopeResourcesHandler`):

```
	/**
	 *
	 * Builds the list of editors that apply to this build that need to be saved
	 *
	 * @param projects the projects involved in this build, used to scope the searching process
	 * @return the list of dirty editors for this launch to save, never null
	 */
	protected IResource[] getScopedDirtyResources(IProject[] projects) {
		HashSet<IResource> dirtyres = new HashSet<>();
		IWorkbenchWindow[] windows = PlatformUI.getWorkbench().getWorkbenchWindows();
		for (IWorkbenchWindow window : windows) {
			IWorkbenchPage[] pages = window.getPages();
			for (IWorkbenchPage page : pages) {
				IEditorPart[] eparts = page.getDirtyEditors();
				for (IEditorPart epart : eparts) {
					IResource resource = epart.getEditorInput().getAdapter(IResource.class);
					if (resource != null) {
						for (IProject project : projects) {
							if (project.equals(resource.getProject())) {
								dirtyres.add(resource);
							}
						}
					}
				}
			}
		}
		return dirtyres.toArray(new IResource[dirtyres.size()]);
	}
```

lwrage:
There's `org.eclipse.ui.ide.IDE.saveAllEditors(IResource[], boolean)`, which is API.

AaronGreenhouse:
Yes, was just about to add that here!

AaronGreenhouse:
Not clear we can directly resume their dialog box... Annoying.

[...]

PR #2232, opened 2020-03-11
Added save dirty editors before instantiation or reinstantiation

AaronGreenhouse:
Fixes Issue #2083


AaronGreenhouse:
> This only saves files that contain the instantiated component implementation(s) but not files that contain packages or property sets that are referenced. At the minimum all modified AADL files used during instantiation should be saved.
> It seems you can also pass in a list of projects to IDE.saveAllEditors such that all modified files (AADL and others) in these projects are proposed for saving. The simplest solution, then, would be to just pass in all AADL projects in the workspace. I think it's not worth the effort to determine the minimal set of files that must be saved, and I see no problem with prompting the user to save more files than necessary.

Yeah, I forgot about this.

Agreed that it is just best to pass in all the open AADL projects.

PR #2242, opened 2020-03-17
JavaFX graphical editor palette implementation

### LINKED ISSUE #2231: Implement JavaFX based palette
Implement JavaFX based palette. Class to be implemented is org.osate.ge.fx.palette.Palette.

- Branch off of the branch for 1183. Once completed the pull request should merge changes back into that branch.
- Should be derived from Region
- Should be driven by the specified PaletteModelController.
- Appearance and behavior of the palette should be similar to the existing OSATE graphical editor palette. However, do not try to replicate the exact look and feel of SWT.
- Palette items which are not contained in a group should be first.
- Palette items should be listed in the order they are provided by the model controller (within a group).
- Palette should *not* contain the collapsing functionality and the top label contained in the current palette. That will be implemented separately.
- Palette should handled being disabled properly. The expected behavior is for the palette to have an appearance that looks disabled and not allow user interaction. This may be handled appropriately by the JavaFX controls being used.
- The current palette will create scrollbars within each group. We want to keep that functionality.

philip-alldredge:
Closes #2231


BrendenMcHan:
@philip-alldredge 
I have found the places where both issues we discussed are occurring.The line of code that causes the latency is: mc.activateItem(ItemArray.indexOf(newValue));I tested to make sure that the way it is finding what index to call is not the issue by doing a system.err.println(ItemArray.indexOf(newValue)); This line does not cause any latency. So, somewhere within the function is what is causing the latency, I think it might be because of the changing back and forth of files to find the correct functions. From my understanding, when the program calls the activateItem function, it goes to the PaletteModelController to call that function, and within that function it has to check that the index is not invalid and in doing so calls the function getNumberOfItems, which in the test case then calls getNumberOfGroups and performs a calculation based on that value that then gets passed back up the same path until it finally makes its way back to the activateItem function. As far as I can tell, checking that the index is not invalid is not necessary as the user cannot click on an item that is not being shown to them, and if it is outside the index, it wouldn't be shown anyway. I may be wrong in that assumption, but would greatly appreciate your input on the topic. 

The second issue, the IndexOutOfBounds exception is being thrown by the line: newValue.setExpanded(true); This line only gets run after the program checks to make sure that newValue is not a leaf, and that it is not already expanded. This exception is only thrown when expanding a branch that is lower on the treeView than the one that is currently selected. When moving up the tree, no exceptions are thrown. I am not sure why an Index related exception is being thrown, as the newValue.setExpanded(true) is not calling an index of any arrays. Again, any insight would be greatly appreciated.

philip-alldredge:
A few things.
- Yes, there is an error with the `TestPaletteModelController.activateItem`. It isn't starting a thread. It's running the code in the current thread. I'll fix that.
- I am unable to run your version because of a missing CSS file. Please correct so I can take a closer look.
- When pushing or referencing code please:
   - Include the code with the problem in your push. Don't comment it out.
   - When including code in comments use the formatting tools to ensure it is readable.
   - I believe you can add comments straight to a particular line of code in order to call attention to it. You can see the changes in the pull request by going to the `Files changed` tab.

philip-alldredge:
@BrendenMcHan sorry on more thing. Can you give me write access to your repository. I can't give you my changes directly.

philip-alldredge:
- I have pushed the fix for the sleep issue.
- I updated the issue with a reminder that we will need to support scrolling within each group.
- I am unable to reliably reproduce the index out of bounds issue. When reporting an issue like this, please describe what I need to do to reproduce it in precise detail.

BrendenMcHan:
@philip-alldredge To recreate the IndexOutOfBounds Exception, select Group 0 or Group1, the one you select should expand. Once it is expanded, select any group that exists below the current group. So if you selected Group 0, select either Group 1 or Group 2, and if you selected Group 1 then select Group 2. This should throw the exception.

philip-alldredge:
Thanks. The issue is somehow related to the collapsing of the previous selected group. Once again, I would recommend moving away from TreeView. There are likely ways to get this to work but it appears to be a road that's going to have issues.

Also, don't use string for the value of tree views.

BrendenMcHan:
@philip-alldredge From what I have been able to tell from researching this issue is that it is thrown when the branch being expanded extends past the boundary of the palette. This would explain why the exception is not thrown when moving up the tree, as the boundary is below where the expanded branch could ever reach. Increasing the height of the palette, which will be done when it is placed inside the container for the graphical editor, should resolve the issue, as the expanding branch will have more room to expand without going outside of the boundary of the palette. 

Also, the main reason for using TreeView is that, at least to my knowledge, there is no other way to implement a hierarchy that has a predefined ways of handling the expansion and contraction of the branches, as well as automatically providing a scrollbar when the items in the branch extend outside the visibility of the user. TreeView also allows for the addition of Icons next to the name of each TreeItem. As far as I know, none of these things are available in any other Layout.

philip-alldredge:
I'm still inclined to think that the indexing is not related to physical sizes but the fact that we are programmatically collapsing a group while handing an event. 

There is nothing magical about the JavaFX tree view.  As far as I know, there is no reason that the the palette can't be implemented as a combination of Button, Label, Image, VBox, ScrollPane, and potential other components. It should be possible to create a Node that represents a group and then use that component within the palette to create a list of groups. That is one way to do it but not the only way.

At this point, I'm not sure if it's even possible to implement the required palette using the tree view. It seems highly doubtful that the desired pinning, scrolling and resizing behavior will be supported.

BrendenMcHan:
The scrolling is already a part of the TreeView, and naturally occurs whenever it is needed. As for pinning, when a group is expanded by clicking the arrow and stead of clicking on the name of the group, it acts as a pin, and the group no longer closes when another group is opened, essentially pinning the group open. 
Ultimately, I will do my best to implement the palette in whatever way you deem the best.

philip-alldredge:
As mentioned in the issue, each group needs it's own scrolling capability. I do not believe that is the normal behavior of a TreeView. Experiment with the current palette to see what I mean.

The behavior you described is not the same as pinning.

I'm okay with using the Tree View if it satisfies the requirements but at the moment it seems 
- Problematic
- Significantly different visually.
- Different functionality wise

BrendenMcHan:
As far as the visual differences go, I am fairly certain I could get the palette to look almost exactly the same using css, however Icons in javafx default to 16x16 which would make each branch's label larger than they currently are. As for the scrolling, I see the difference you mentioned in the current palette, and I will look into possible ways to recreate that in javafx.

philip-alldredge:
I think the graphical editor's current icons are 16 x 16 so that shouldn't be an issue. Update the test model controller to match then. However, there shouldn't be anything stopping us from displaying larger images or scaling the specified image.

BrendenMcHan:
@philip-alldredge After doing a lot of research into the Group and ScrollPane documentation for JavaFX, I think the best way to get the scroll functionality and a closer behavior than what is possible with the TreeView is to use Labels or possibly Buttons within a root Group that would display the hierarchy, and when each label is pressed it creates a ScrollPane within that Label/Button. That ScrollPane would contain the components of each Group.

philip-alldredge:
That is similar to what I was suggesting. Keep in mind that you shouldn't create and destroy the scrollpane each time it is toggle. Just hide it(potentially with an animation).

BrendenMcHan:
I think I might be able to create all of the ScrollPanes on initialization, and then remove the current ScollPane from the screen before making a new one visible.

BrendenMcHan:
It may also be looking into using the MenuButton class. This class opens a dropdown menu when clicked that may be similar to what we are looking for.

philip-alldredge:
FYI. A node can be hidden and removed from layout using the visible and managed properties.

https://stackoverflow.com/questions/28558165/javafx-setvisible-doesnt-hide-the-element

philip-alldredge:
I would suggest breaking the implementation into multiple parts. One class would be a node that represents a single group. At that point, it should be easy to instantiate the node for each group.

BrendenMcHan:
That could prove incredibly valuable to the ScrollPane implementation idea.

BrendenMcHan:
@philip-alldredge 
As I was working on implementing the palette using scroll panes, groups, buttons, and Vboxs as we discussed yesterday, I ran across a problem regarding the expected functionality. Because each group must be made dynamically using a for loop to support variable number of groups, no group can access the properties of another group. Essentially, each group thinks it is alone, only the Vbox that is each groups parent knows that any other group exists. Because of this, there is no way, that I know of, to set each button to close the scollpanes controlled by other buttons. Each group already has an event listener that opens and closes its own scollpane when clicked, however, because each group is created in a for loop, and therefore has the same name and all the same properties, they cannot be referenced after the end of the for loop. If you have any suggestions as to how we could work around this issue, they would be greatly appreciated. 
I have pushed my updated code so that you may take a deeper look into what I mean.

philip-alldredge:
Okay. I'll take a look.

philip-alldredge:
@BrendenMcHan I pushed some changes to your branch. I added a PaletteGroup class for you to implement. I added "TODO" comments to Palette and PaletteGroup with what I believe is the best way to proceed.

The general idea is to break things into two pieces. Palette Groups and the Palette itself. Each palette group manages it's own expansion state but provides a JavaFX property that allows the palette to listen to changes and adjust it. The palette keeps track of the group it contains. 

My TODOs indicate that it should keep a list of groups. That may not be strictly necessary if the palette only contains groups. In that case, the result of getChildren() should be a list of groups. However, maintaining a separate list is likely cleaner and less likely to break as we expand the palette.

Please let me know if you have any questions or concerns.

BrendenMcHan:
@philip-alldredge 
I noticed that in the PaletteGroup class constructor, you included the integer value groupIndex. However, this value is not used within the class. My question is what is the intended use of passing in this value to the constructor?

philip-alldredge:
It is required to get the icon, label, and items in the group. The group node should display the contents of a single group. In your loop you will create an instance for each group.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteGroup.java
```diff
@@ -0,0 +1,60 @@
...
+		groupButton.setGraphic(new ImageView(mc.getGroupIcon(groupIndex)));
+
+		buttonBox.getChildren().add(groupButton);
+
+		ScrollPane scrollPane = new ScrollPane();
+		for (int i = 0; i < mc.getNumberOfItems(); i++) {
+
+			if (mc.getItemGroup(i) != null && mc.getItemGroup(i) == groupIndex) {
+
+				Button itemButton = new Button(mc.getItemLabel(i));
```
philip-alldredge: Create a PaletteItem class which is a region that represents an item. Pass in the PaletteModelController and the item index to its constructor. It will be needed because the item will be needed at the Palette level as well to show items which are not in a group. I had forgotten about that when writing earlier notes. 

Let's discuss if you have an alternative idea.

philip-alldredge:
There current WIP implementation requires looping through each item for each group. This will likely not be a performance issue but we should consider adjusting the PaletteModelController to avoid needing to do that. This can be revisited once everything is functioning.

BrendenMcHan:
@philip-alldredge I agree that avoiding looping would be much easier, as statically assigning each group and item within each group would make referencing them much easier. As I have been working on the palette since we moved to scroll pane instead of tree view, I have not been able to figure out a way to index the scroll pane contained by each palette group. It cannot be referenced by using getChildren(), as palette group is not a node, it is a region, and it also cannot be referenced using dot notation because the scroll pane is not a property of the class, it is just an item in the class.
However, I believe the reason we started with loops is because we won't know how many groups or items the user will have implemented because of the use of other various plugins, but being able to avoid the loops would make the process much easier.

philip-alldredge:
- I was not saying we should define the items statically.  My only concern with the loops is that the general structure is not efficient. If we have n groups, the loop will be executed n times.  It's a minor detail and unlikely to have a measurable impact. The last note was primarily for me and something that should be easy to cleanup when functionality is achieved.
- Palette group is a node. A Region is a type of node. See: https://docs.oracle.com/javase/8/javafx/api/javafx/scene/layout/Region.html
- You shouldn't try to access the palette groups' scroll pane directly. It's generally a bad idea to allow other classes to access the internals of an object directly. It breaks encapsulation. The user(Palette) of the PaletteGroup class shouldn't need to know the details of the PaletteGroup class.
- For the PaletteGroup class, the expanded property should control whether the group is expanded/the scroll pane is visible.  Lookup how JavaFX properties work if needed.
- Do not access the scrollpanes directly from the Palette. However, if one needed to do such a thing, it would just be a matter of storing the ScrollPane in a public field instead of a local variable.  However, that isn't recommended. It would be better to store it in a private field and then access it as part of methods provided by the containing class such as setExpanded. That likely won't be necessary when using the expanded property but this explanation is just provided as general information.

BrendenMcHan:
@philip-alldredge I have been trying to get the palette to appear on screen with the new scroll pane implementation, but have been unable to find out why it won't show up. In theory, it is just a few nested Vbox's and the code I have looks like it should work. If you could take a look at it and see if there is anything obvious I am overlooking, that would be greatly appreciated.

philip-alldredge:
@BrendenMcHan nothing is getting added to the palette group. The "buttonBox" VBox is never added to the palette group itself. Something like `this.getChildren().add(buttonBox);` should fix it.

philip-alldredge:
@BrendenMcHan FYI - I'm considering some changes to the ModelController, etc that would cause some minor disruption. I just wanted to make you aware. I'll make the changes this afternoon at the earliest so be sure to push your changes. That way, you can pull in the morning and there won't be any merge conflicts.

BrendenMcHan:
Ok, will do. Thanks for letting me know.

philip-alldredge:
@BrendenMcHan I have pushed some changes. Before I started making changes, I tried to run the palette test app and nothing was displayed. Notable changes:

- Renamed `PaletteModelController` to `PaletteModel`. This was done to follow the naming convention that is evolving as part of other work being done.
- Reworked `PaletteModel` so that there isn't a need to keep track of indices and it is now easier to determine which items are in a group or at the root level. 
- Added basic javadoc to the `PaletteModel`.
- Adjusted `PaletteModel` to provide a more type-safe API. If you are unfamiliar with Java generics, please see https://docs.oracle.com/javase/tutorial/java/generics/index.html and https://en.wikipedia.org/wiki/Generics_in_Java
- Removed layoutChildren(). I originally added it just in case it was needed and it doesn't sound like you expect to use it.

I also noted that there was some ugly code formatting. Are you editing using Eclipse? The project should be setup to automatically format things when you save.

BrendenMcHan:
@philip-alldredge Yes, I am using eclipse to edit, and thank you for letting me know about those changes, they seem like they're going to be very helpful.

philip-alldredge:
@BrendenMcHan a palette group should not reference any other palette groups. It should use the expanded property to notify any listeners of a change in its expansion state. The `Palette` should listen to the expanded state of all palette groups and update the expansion of the other groups.

BrendenMcHan:
@philip-alldredge Since the expanded state of all groups need to be changed when a group is selected, wouldn't the change have to happen within the setOnAction method of each group Button?You could also possibly do it by having a function specifically used for closing all groups except the one that is clicked, and then you could call that function in every group Button setOnAction method.

philip-alldredge:
Yes, the expanded state of all unpinned groups will need to change when a group is expanded. 

No, it doesn't need to be in the `PaletteGroup`. The only thing the on action should do is toggle the value of the expanded property. `PaletteGroup` doesn't have easy access to any other groups and it shouldn't have access to them. Just like a regular button doesn't access all the other buttons in your application, the palette group shouldn't interact with other groups. The `Palette` owns all the buttons and its responsibility to communicate with them. In this case, it communicates with them by setting the expanded property/getExpanded/setExpanded.

- In the action function of the group's toggle button, toggle the expanded property.
- In `PaletteGroup`, ensure the items(scroll pane) are shown and hidden based on the expanded property value.
- In `Palette`, listen to changes in the expansion state. When a group is expanded, set the expanded property of all other groups to false.

BrendenMcHan:
@philip-alldredge I have an idea about how I would achieve that functionality within palette, but I wanted to run it by you first to see if there was a more efficient way. I think you could achieve it by adding a listener onto each button that is created and when one of them is pushed it notifies the palette and then the palette would iterate through an array of the other groups and close them.

philip-alldredge:
@BrendenMcHan how does your approach differ from what I've been saying? I want to make sure I understand what you are proposing.

BrendenMcHan:
@philip-alldredge I don't think that it does, I was just making sure I had a proper grasp on how to implement the ideas you suggested.

philip-alldredge:
Okay, in that case it sounds about correct. The palette should listen to changes in the expanded property. The group shouldn't access the palette or other groups directly.   I would do it like described in the bullets of my previous comments. Let me know when you have something and I can confirm that you are going down the right track.

BrendenMcHan:
@philip-alldredge I think I am on the right track, but I am having trouble finding a way to have the listener skip over the group that was selected when it iterates through the list closing all of them. I have pushed my new code to the branch so that you can view it and see what I mean. The code is commented in the Palette.java file

philip-alldredge:
@BrendenMcHan a few thoughts.
- Just compare the groups using !=. You are already creating a new listener for each group so that listener can access the group from your outer loop. You may have to add the `final` keyword to your palette group. For example `final PaletteGroup<G, I> paletteGroup = new PaletteGroup<>(model, group)`.
- In this case, I don't believe there is a need to loop through groups using index. Use a for-each loop whenever possible as it is more clear. https://www.geeksforgeeks.org/for-each-loop-in-java/ . I understand your intention was to compare by index. If that was indeed needed, the for loop you had would have been a good approach.

Let me know if that makes sense.

BrendenMcHan:
Thank you, that information really helped, I am going to do another push so you can make sure that the closing functionality matches your expectations.

philip-alldredge:
The palette groups appear to be expanding as we discussed.

I realize that I'm making comments while you are still working on things, but I want to make sure you are aware of a couple things,
- Ensure all variables match standard java variable naming conventions. Specifically, all variables should be lower camel case. https://en.wikipedia.org/wiki/Camel_case  . Single upper case letters are usually used for type parameters (G and I in this case). 
- All warnings will need to be resolved before finishing.

What are your planned next steps? Personally, I would fix warnings and implement the item activation. After that I would focus on the layout and appearance of things.

BrendenMcHan:
@philip-alldredge My next steps will probably be to go back through and change any variable names that don't follow the lower camel case naming convention, an then to go through and fix the warnings and implement the item activation as you mentioned, and then move on to the layout and appearance.

philip-alldredge:
Great

BrendenMcHan:
@philip-alldredge Currently, I am working on getting the group and item buttons to fill the width of the current window dynamically, so that if the palette is resized the groups and items will stretch or shrink to fit the window they are given. Currently, I have set each scroll Pane to automatically size its components based on the width of the scroll Pane, this will ensure that all items will be the same size, and will fill the entire space available. I have also done the same thing with the VBox that houses each group. In theory, the next step would be to bind the width of the VBox that houses the entire palette to the current width of the available window. However, since the window and stage are created elsewhere in the file, I am unsure how to bind the width property of the palette VBox to the width of the current window, and was wondering if you had any suggestions on how to achieve that, or if the strategy of binding the width properties is even the right course of action in this situation.

philip-alldredge:
@BrendenMcHan you can modify `NodeApplication` to call `setFillWidth` on `container` if that will help. Please push your code so I can take a look at the issue.

philip-alldredge:
On the currently pushed version, the palette fills the window but its contents does not. This is without modifying `NodeApplication`. I confirmed this by setting the background color of the Palette.  The palette should not try to bind anything to size of the container.

BrendenMcHan:
@philip-alldredge I just pushed my code so that you can view it.

philip-alldredge:
@BrendenMcHan I pushed a change to help show the problem. The palette is growing but the `paletteVbox` is not. I suggest you look at https://docs.oracle.com/javase/8/javafx/api/javafx/scene/layout/Region.html to perform a layout for the palette VBox.

> To implement a more custom layout, a Region subclass must override computePrefWidth, computePrefHeight, and layoutChildren. Note that layoutChildren is called automatically by the scene graph while executing a top-down layout pass and it should not be invoked directly by the region subclass.

I would recommend doing that for `Palette` and `PaletteGroup` so they are sized appropriately. The reason we don't want to just inherit from `VBox` or other pane is that we don't want to expose and editable children list to users of these classes.

BrendenMcHan:
@philip-alldredge Thank you for all your help.

BrendenMcHan:
@philip-alldredge I am working on the resizing of the palette and paletteGroup respectively, and I understand that I need to override their prefWidth properties, but I am unsure as to what they should be set to in order to fill the entire window.

philip-alldredge:
@BrendenMcHan I believe classes for which you are implementing the custom layout only contain a single child so it shouldn't be too difficult. `Palette` already takes up the entire window. Its `layoutChildren` just needs to be implemented to set the child's size to match its size. The same can be done for `PaletteGroup`. `computePrefWidth` and `computePrefHeight` should just likely just call the `prefWidth` and `prefHeight` of the child.

BrendenMcHan:
@philip-alldredge I have just pushed an updated form of the code that has resized all items properly and also made a small change to the number of items in the TestPaletteModel so that I could test the scrolling functionality. I did, however, notice one small issue with the scroll functionality. The scrolling functionality does not kick in until the items within the scrollPane take up all the space that the palette is given which, if I remember correctly, is not how the current palette functions. What also ends up happening because of this is that the group buttons above and below the group that is expanded get squished into, and sometimes overlap with each other and create a mess. 
As far as the problem of when the scrolling functionality kicks in, I am not sure if there is a way to change how much of the screen has to be taken up before it will kick in. 
In reference to the items being pushed off screen and being squished into each other, I think it would be possible to resolve this issue by nesting the entire palette into a scrollPane and allowing the items within it to go outside the current viewing window. This should allow the Items above and below the expanded item to exit the screen without being squished into each other or overlapping.

philip-alldredge:
@BrendenMcHan 
I've taken a look at things and a closer loook at the current palette. See below for a list of things that I think need to be improved. It also contains some changes from what I had previously told you.

- Do not implement pinning functionality.
- Allow expanding multiple groups. Do not collapse other groups when a group is expanded. Sorry for the confusion. I looked at the current palette and it's behavior appears like it would be difficult to implement and not necessarily something worth replicating.
- Don't worry about the squishing when the window is too small. You are correct, we'll likely put it in a scrollpane to handle that case. However, we may need to ensure minimum sizes are appropriate.
- Icons and text need to be left aligned. 
- Groups need to be visually distinct.
- Item icons and text needs to be indented.
- Activated item indicator needs to be more distinct. It was hard to notice it.


Does this answer your questions?

BrendenMcHan:
@philip-alldredge Let me make sure I understand a few of the things from your list. In regard to the expanding groups, I have already implemented the functionality where all other groups close when a new group is expanded. Did you mean that I should not implement that functionality if I had not already done so, or that I should remove that functionality. 
Also, when you say that the groups need to be visually distinct, do you mean they should all look different from each other, or that they should look different than the items they hold?
Lastly, would you like me to simply print out something to the console when an item is activated just for testing purposes, or would you like me to have the item that is selected change color to indicate that it has been selected?

philip-alldredge:
- You should remove the closing of other groups when a group is selected. The other palette did *not* behave that way. It closed unpinned groups whenever it ran out of space. That's a bit complicated to implement and I don't think that's necessary. It would be better to just avoid automatically closing groups. Because of that, you don't need pinning functionality either.
- Groups should be visually distinct from items. Right now, they look the same and it's hard to view the palette. 
- No need to print anything out. 
- I previously mentioned not doing too much CSS but I'd like for you to experiment and see how much CSS/styling it is to make the palette look like the current one.

BrendenMcHan:
Got it, thank you for the clarification. Also, I have run across something strange when trying to align the text within the buttons. I've been using setTextAlignment to move the text, and setContentDisplay to move the Icon, however, neither function has any affect on the buttons. I tried setting the alignment for both the Palette Items and the Groups and nothing worked. Is there another way of setting the text alignment that I am missing?

philip-alldredge:
I'm not sure. Please look into it and let me know tomorrow if you haven't been able to find the answer.

BrendenMcHan:
@philip-alldredge So I have come across something rather interesting. I have tried re-aligning the text on each button from each file using the text alignment function, and have tried using css as well to properly place the text, however, nothing works. It seems that for some reason, the text on the buttons is locked into the center. I have one idea of why that might be, but I am not sure if it is relevant. I think it is possible that javafx does not recognize the text as actual text, but sees it as a label and does not know the contents. This would only be true if a label was used to create each button at compile time.

philip-alldredge:
@BrendenMcHan I just looked through the docs and it appears you missed the alignment property. A quick test shows that it will do what you need. Found it by searching for `alignment` on the javadoc. Use `setAlignment`. https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Labeled.html#setAlignment-javafx.geometry.Pos-

I appreciate the effort you put into looking for the solution but it's important to read the documentation closely. The textAlignment property explicitly states it only affects multi-line text.

BrendenMcHan:
@philip-alldredge I have just pushed an update to the code where I have styled the palette to closely resemble the current palette. I also managed to do all the styling necessary without creating a separate css file.

philip-alldredge:
Looks pretty good! There are a few things I noticed that will need to be fixed in order to finish up the component.

- Look into using a toggle button for the group button.  I believe this will give it a different appearance when it is selected. Then you can remove the `expanded` field and update all methods that use it to use the selected state of the button. 
- It doesn't look like the highlght color on the items exactly matches the current one. Might as well make it match exactly.
- Indent the items like in the current palette. (This was previously mentioned but I assume it was missed).
- Try to make the group buttons look and behave like the current one. This includes the mouse over highlight behavior. You did it for the items so it looks like it is doable for the group buttons as well.
- Run the code formatter on the code. There are a lot of odd new lines in it.  Right Click -> Source -> Format.

BrendenMcHan:
@philip-alldredge The color was one of the things I was most worried about, my computer is not the best when it comes to color resolution so it is very hard to tell the difference in shades of the same color. I didn't miss when you mentioned indenting the items in the palette, I have just been wrestling with how to achieve it, since there is not a position value that can achieve it. You could simply indent the text by putting space before the label, however that leaves the icon bound to the left side, creating an awkward gap.

philip-alldredge:
Okay, I'll let you keep working on the intention. For the color, I'd recommend taking a screenshot and checking the color in an image editor such as Paint.

BrendenMcHan:
@philip-alldredge When you get the chance, can you check the new colors of the palette to see if they match? Thanks.

BrendenMcHan:
@philip-alldredge I just pushed an update to the code with the proper indentation of the palette items. In order to keep groupless items from being indented, I added a boolean parameter to the PaletteItem constructor. This boolean contains whether or not the item has a group. When the constructor is called inside the for loop that handles the creation of the groupless items, it is set to false, and when it is called on items from within the PaletteGroup class, it is set to true.

philip-alldredge:
- Palette item highlght color stiill does not match. Did not check others.
	- The likely reason that you are not getting the colors you want is because the color being set had an alpha value of 0.2. The alpha value affects how the color blends with other colors. Since we are just trying to set a specific color, we want an alpha value of 1.0. (https://en.wikipedia.org/wiki/Alpha_compositing)
	- For example, I believe the color for the highlighted palette item should be: rgba(252,228,179,1.0)
- Root items aren't indented like they are in the current palette.
- Use constants like shown in https://stackoverflow.com/a/46525284 rather than having inline styles. It'll be either to review and customize.
- Git rid of all the blank lines at the start and end of every block.
- Some borders or something similiar are needed in some of the components. For example: compare how palette groups look between the two palettes. The colors on the new one blend together in some cases because they don't have any sort of border separating them.

philip-alldredge:
I don't think that boolean is needed since I believe in the current palette, item contents are always indented. Please confirm.

BrendenMcHan:
@philip-alldredge I believe I have resolved all the issues you brought up, I also added a small amount of space between the group buttons to more closely resemble the current palette.

philip-alldredge:
@BrendenMcHan looks like things are moving along. I'd like you to keep refining the UI and then start working on improving the integration.
		
- Continue refining the palette UI
	- Something looks off on the palette groups. Can you check.
		- Color/Gradient. Default, hover, and selected.
		- Font sizes
		- Top and bottom padding for group labels(inside the button).
	- Palette Items. Something looks a bit off. Check top and bottom borders.				
	- Borders
		- Remove borders on the left and right of the buttons and groups. It looks odd because some parts have borders and some don't. The container of the palette will have a border when integrated to the editor.
	- Having the borders on the side of the groups  Just need them between groups. Horizontal. Don't think we need the vertical borders as well.
		Avoid the one below the last open group.
	
- Next, start improving the integration into `org.osate.ge.gef.ui.editor.AgeEditor`. This editor can be opened by right clicking on a diagram file and selecting Open With->AADL Diagram Editor (GEF)
	- Instead of adding the palette to the split pane, add a container which contains a "Palette" label, a collapse button and a scrollpane with the actual palette. When finished, it should resemble the current palette.
	- Add a border to the entire palette + label + etc container
	- Look into the "proper" way to set the initial divider position.

BrendenMcHan:
@philip-alldredge I have just pushed an update that takes away the side borders from the groups, but something I noticed while trying to fix the issue is that there will always be a natural border where the palette ends. From what I have read, a lot of border related things do not work in javafx css like they do in html. Styles like border-style and border-left, border-right, border-top, and border-bottom don't function at all in some circumstances, and border-style behaves very differently than it does in html. It shouldn't be a problem in this instance, I just wanted to make you aware if you weren't already because it may pose problems with other areas further in the future.

Also as a side note, I noticed that when you select an item in the palette, there is a three second time period where highlighting that item will not change it's color, it remains blue as it should. However, after those three seconds, the item can be highlighted in orange and loses its blue color, which would cause the user to think that the item is no longer selected. I am fairly certain that this is caused by the opening of a new thread in the TestPaletteModel, and would not happen in the actual implementation of the palette, but again, I just wanted to make you aware of it so you not be confused if you ran across the same thing during your review of the code.

BrendenMcHan:
@philip-alldredge Should I be making the changes in the integration to the Palette.java file, or to the AgeEditor.java file?

philip-alldredge:
Integration changes should not be made to the `Palete` itself. It should be in `AgeEditor`

BrendenMcHan:
@philip-alldredge For the past two days I have been working on properly resizing the palette to match the layout of the current palette, and I have run across some very strange issues. I have tried setting the minimum and preferred width of every node from the palette to the palette Container and every layer in between, and the results seem random. I can set the prefWidth to 100 and it changes nothing, but changing it to 1000 makes it fill the screen. I thought that maybe the minWidth was getting in the way of making it smaller so I changed the minWidth to 10, and again nothing changed. Because the palette was taking up so much space I tried setting the maxWidth of the paletteContainer to 100 to keep it from being too large, and instead of fitting that width, the palette resizes so small you can barely see it. I raised the maxWidth in stages of 100 all the way to 1000, while keeping the prefWidth also at 1000 and the minWidth at 10, and the entire time the palette stayed at a minuscule size.

 From what I have been able to gather from research, the way that the initial divider position is being set is correct, however, it doesn't seem to make a difference because the divider position is always overwritten by the prefWidth, maxWidth, and minWidth values of whatever is contained in the splitPane. I hate to come to you with a problem like this because it should be very simple and I am sure I'm probably missing something quite small that is throwing me off, but I don't know what else to try.

BrendenMcHan:
@philip-alldredge I have found the issue mentioned in the comment above, the paletteContainer must be given a maxWidth otherwise it will take up all available space, but after setting a maxWidth, the prefWidth property is used to determine the actual size, by going through that process and binding the scrollPane widthProperty to the widthProperty of the paletteContainer, I was able to achieve the required size and layout.

philip-alldredge:
Glad you got it worked out. I took a quick look and it looks like you are making good progress. I noticed however, that the phone sizes are much larger than in the current palette. It's very clear when swapping back and forth between the editor windows. Please take a look at that.

I also noticed that clicking multiple items turns multiple of them blue. There is only one active item. The palette doesn't seem to be listening to changes quite correctly. Please look into that as well.

BrendenMcHan:
@philip-alldredge I have just pushed an update with a small modification to the font sizes. I originally had them set at 15, which was too large, and I have now set them at 14. To me, the text now looks smaller than in the original palette, but it may just be the fact that it is different words that is throwing me off, so I wanted to get a second opinion.

BrendenMcHan:
@philip-alldredge I have just pushed an update which adds the collapse button functionality and styling for the title area. The collapse button does not have a graphic yet, so I implemented a permanent border around it to make it visible until the graphic is added. Also, I looked into the issue of multiple items being highlighted blue even though only one is selected. The item that stays blue after another item is no longer selected knows that it has been deactivated, but cannot update the background color until the next time it is hovered over. The item buttons do not have access to each other, and that makes the situation a bit trickier. However, I think by changing the item buttons to toggle buttons and placing them into a group with each other should fix the problem, since only one toggle button from a group can be selected at once.

philip-alldredge:
@BrendenMcHan as for the font size. Please take a screenshot and measure to compare. WIth regards to the active item indicator... your comment encouraged me to look at the code.  You *should not* be setting the style anytime you call `activateItem`. You should listen to changes in the property and update whether something is highlighted whenever that changes. It is not guaranteed that `activateItem` will activate the item.  It is only a request and there may be cases where the action is not successful.

BrendenMcHan:
Since the items created are item buttons, not palette items, I was not able to find a way to call the activeItemProperty on them to be able to listen to changes in that property.

philip-alldredge:
@BrendenMcHan please google and readup about how JavaFX properties work and how to listen for them and avoid memory leaks.

BrendenMcHan:
@philip-alldredge I have pushed an update to the code with my attempt at listening to changes in the activeItemProperty to change the style of an item. It does not work as intended, multiple items are still able to be selected despite my best effort, and occasionally when a new item is selected all old items become selected again. I am not sure why what I did didn't work, and I would greatly appreciate if you could look into the code and see where I went wrong.

philip-alldredge:
@BrendenMcHan From looking at the current code, it looks like for each palette item you are foreaching through the "toggles". The toggles in this case are just 1 item so it makes the code confusing. It doesn't make sense to create a toggle group with just one item. Please remove. 

For the actual problem.. you are setting the style based on the state of the toggle button. Buttons in a group by themselves will not be untoggled when you select another one. The toggle state and/or the style should be set based on the selection coming from the model. The model should the authority of what is selected, not the state of the toggle buttons. The active item may be changed in other ways that not part of the palette.

Additionally.
- All fields should be private in your classes. Make them final as well if they aren't reassigned after initialization.

philip-alldredge:
@BrendenMcHan Also, please use WeakChangeListener to prevent memory leaks. A memory leak could occur if the model with the property isn't freed when the palette is closed. While this may be true, it would be best not to rely on that behavior. Lookup how to use `WeakChangeListener`, when using it, you have to hold on to a reference to your listener.

BrendenMcHan:
@philip-alldredge The reason I was trying to change the style based on the state of the toggle button is because in the java documentation, it says that toggle buttons placed into a toggle group together cannot be selected at the same time. Only one at a time can be selected. Based on that, I was trying to add them all to a toggle group as they were created, and then add a listener to the selection state of each button. Unfortunately, because the palette Items are generated dynamically, I wasnt able to find a way to make that happen without giving undue access to the palette item class. I would have had to declare the toggle group inside the palette.java file, but implement it only for the items.
Also could you clarify what you meant by saying the selection should come from the model. Thanks

philip-alldredge:
@BrendenMcHan

The button for the active item is the one that needs to be shown as "selected". By attempting to set th style based on the toggle state of the button, the active item property is being ignored. The palette should show the active item regardless of why the active item changes. From the palette's perspective, the active item could change at any time. 

In our test example, the active item will change to the requested item and then reset itself after a few seconds. If the toggle state is checked, then the indicated active item will never be updated. The palette item should not assume anything about the active item. It should not even assume the active item is changed after `activateItem` is called. 

Additionally, setting the style in multiple places may cause some unexpected issues. For example what happens if the tool deactivates while you are already hovering over the item? I believe it will not be shown as hovered.

I would recommend:
- Track whether the mouse is hovering over the item in a boolean flag. Do not check whether item is active.
- Create a `updateStyle()` method that will check the active item from the model and the boolean flag described above and set the style to the appropriate value. 
- Call `updateStyle()` at the end of your mouse exited and mouse entered event listeners. 
- Update the active item property listener to call `updateStyle`.
- Do not set style anywhere else. 

By doing what is described above, then you can be sure that the style will always be set to the same value under the same conditions regardless of event ordering.

BrendenMcHan:
Thank you so much for the advice. It really helps a lot.

BrendenMcHan:
@philip-alldredge I have been trying to implement the changes you brought up in your last comment, and I have run across a few issues. I've been having a hard time figuring out how to reference the specific item button based on the item that is active, since as far as I can tell, there is nothing that ties them together, by that I mean I cannot call the button based on the item. I am also having a hard time figuring out how to restyle the button that was active before the new item was selected. I tried using an arraylist containing all the item buttons, and then iterating though them in a for each loop, which would work if you changed the style based on the toggle state of the buttons, but since the active item can change in other ways, this approach wont work, and as I stated above I have been unable to find a way to call an item button based on the item is corresponds to. I will push an update with the methods I have tried so far.

philip-alldredge:
@BrendenMcHan the `PaletteItem` should know what item it is. That is passed into the constructor.  You'll need to compare its item with the active item from the model to determine if it is active.

BrendenMcHan:
I am already comparing the item with the active item during mouse events and during the on action method when each item is selected. The problem I am coming across is how to iterate through all other items to deactivate them and change their style back to the idle item style.

philip-alldredge:
@BrendenMcHan as described above, you still need to have a listener for the active item property that calls `updateStyle`. Your `updateStyle` method will need a third case to handle when it's not active and not hovering.

A couple other things I noticed:
- `updateStyle` shouldn't have parameters. You can just store the information you need in fields to avoid needing to always pass the same arguments. 
- Don't use an array to store the hovered state. I am guessing this was a workaround when you had it is as a non-final local variable. It's not needed when it's a field.

BrendenMcHan:
What is a field in javafx, I have tried searching for it, but all that shows up is textFields.

philip-alldredge:
Fields are not a JavaFX concept. It's a language concept. In C++ they are often called member variables. In Java, C#, Rust, and likely others, they are called 

http://tutorials.jenkov.com/java/fields.html

BrendenMcHan:
I was able to make the hovered state a field and remove the array, however, since paletteModel and Item are used in the class Constructor, making them a field throws an error.

philip-alldredge:
@BrendenMcHan create a field and assign it. See: https://github.com/osate/osate2/blob/master/ge/org.osate.ge.swt/src/org/osate/ge/swt/prototypes/PrototypeEditor.java#L68 for a rough example.

BrendenMcHan:
@philip-alldredge I have made all the appropriate fields after viewing your example, which proved to be very helpful. The only problem I am having now is with the update style method. Inside the update style method, I am looping through all the items by looping through the groups they belong to first. I am trying to iterate through the groups by doing "G group : model.getGroups()", however it does not recognize G as a type. I assume this is because G is declared in another file. So my question is what I need to change in order to get my current method to work, or if there is a better way of iterating through the groups. I will push an update so that you can see the problem in context.

philip-alldredge:
@BrendenMcHan you should not be looping through groups in your updateStyle method. "G" is used in other places as a type parameter. You could define a type parameter if you needed to reference it, but you do not. PaletteItems should not care what group they are in. 

Below is a list of changes you should make. Most of this was already covered in previous discussions but I've repeated it to consolidate things. As a high level overview: the idea is that the `PaletteItem` will update it's style based on it's current state and the state of the model. That means whenever it's internal state changes(hovered status) or the model state changes(active item property), `updateStyle` should be called. Additionally, `updateStyle` should be used to initialize the style. Doing these things ensures that the style of the `PaletteItem` will always be the same given the same state. That's good because it prevents the code from relying on specific ordering of events and prevents odd bugs.

- `updateStyle` should only update itselfl. It should set it's style based on whether is is active, hovering, etc.
- `updateStyle` should update it's style regardless. There should always be exactly one call to `setStyle` within it.
- `updateStyle` should be private
- `updateStyle` should not be called as part of the onAction. (Palette item isn't changing state).
- `updateStyle` should be called as part of a listener for the activeItem property.
- The button's style should not be set anywhere other than `updateStyle`.
- Call `updateStyle` at the end of the `PaletteItem` constructor to initialize the style.
- Remove `setItemHovered` unless there is a reason for it to exist. If there was a reason for it to exist, it should be private. Nothing else should be modifying the `PaletteItem`'s state. In general, everything should be private unless there is a need for it to be otherwise.
- Remove `isItemHovered`. 
- Recommend removing the `item` prefix on your fields. Everything in the class is related to the palette item.

BrendenMcHan:
You're right, paletteItems do not care what group they are in, but that was the only way I could think of to loop through all paletteItems to deactivate them when another item is selected. Also, when onAction is called the state of the item is changed to selected, so updateStyle would need to be called right?

philip-alldredge:
@BrendenMcHan Yes and No. Yes: the style of an item needs to be changed when it is activated or deactivated. The palette item needs to update itself.

No: when onAction is called the palette item requests that the item is activated. The item may not be activated. The item may be activated later.  It may be activated outside of the palette. It will definitely be deactivated outside the palette. The model provides an activated item property which can be used to listen to changes to the active item. Each palette item should use that.

BrendenMcHan:
@RyanMcilnay I have just pushed an update to the code with my best attempt at adding a weak change listener to the activeItemProperty. From what I can tell everything is mostly right, but I am coming across an error that I can't seem to work around. When adding the change listener in the PaletteItem.java file, it throws this error: "An anonymous class cannot subclass the final class WeakChangeListener<ReadOnlyProperty<I>>"

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,78 @@
...
+		Button.setFont(new Font(14));
+		Button.setOnMouseEntered(e -> {
+			Hovered = true;
+			updateStyle();
+		});
+		Button.setOnMouseExited(e -> {
+			Hovered = false;
+			updateStyle();
+		});
+		model.activeItemProperty().addListener(new WeakChangeListener<ReadOnlyProperty<I>>() {
```
RyanMcilnay, line 40: change this to
````
model.activeItemProperty().addListener(new WeakChangeListener<I>(this::resourceChangeListenerHandler));
````

File: ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+
+		Button.setAlignment(Pos.BASELINE_LEFT);
+		Button.setGraphic(new ImageView(model.getItemIcon(item)));
+		Button.setOnAction(e -> {
+			model.activateItem(item);
+		});
+		this.getChildren().add(Button);
+		updateStyle();
+	}
+
```
RyanMcilnay, line 55: Add this method here
````
    private void resourceChangeListenerHandler(ObservableValue<? extends I> observable, I oldValue, I value) {
	System.err.println(observable + " observable");
	System.err.println(oldValue + " oldValue");
        System.err.println(value + " value");
    }
````

**Comment by BrendenMcHan** [comment]
@RyanMcilnay I have made the changes you suggested, and it seems like the listener is working, however, the functionality isn't quite right. When items are deactivated, they do not automatically update. Also, I noticed the lines of code in the resourceChangeListenerHandler meant to print out the values of each value, so I have included the output produced by selecting an item below.

ReadOnlyObjectProperty [value: org.osate.ge.fx.palette.TestPaletteItem@2b9b7ea5] observable
null oldValue
org.osate.ge.fx.palette.TestPaletteItem@2b9b7ea5 value

**Comment by BrendenMcHan** [comment]
Disregard my comment about the functionality, I added a call to the updateStyle method within the resourceChangeListenerHandler, and the functionality matches what is expected. Thanks for all your help.

**Comment by philip-alldredge** [comment]
@BrendenMcHan I still have some comments about the appearance. See below. I'll open a review about code related issues. 
- Collapse button appearance. It just appears as a box with a dotted outline. It should look similar to the current one.				
- Sizing isn't working properly/not useable. See attached screenshot. Some groups are so small that the user can't see the scrollbar/less than a single item. Something needs to be done so that they share the space.		
- All the fonts are still significantly larger than the current ones. See comparison screenshot

Sizing

Sizes

File: ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteGroup.java [hunk 4]
```diff
@@ -0,0 +1,89 @@
...
+import javafx.scene.layout.Region;
+import javafx.scene.layout.VBox;
+import javafx.scene.text.Font;
+
+/*
+ * Represents a single group in the palette.
+ *
+ */
+class PaletteGroup<G, I> extends Region {
+	final VBox buttonBox = new VBox();
```
philip-alldredge, line 17: All these fields should be private.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteGroup.java
```diff
@@ -0,0 +1,89 @@
...
+	final ToggleButton groupButton;
+
+	final String IDLE_GROUP_STYLE = "-fx-background-color: linear-gradient(rgb(247,247,247),rgb(200,200,200));"
```
philip-alldredge, line 20: These constants should be private and static. Please place them at the top of the class.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,79 @@
...
+import javafx.scene.layout.Region;
+import javafx.scene.text.Font;
+
+class PaletteItem<I> extends Region {
+
+	private final Button Button;
+	private boolean Hovered;
+	private final PaletteModel<?, ?> model;
+	private final I item;
+	private static final String HOVER_ITEM_STYLE = "-fx-background-color: rgba(252,228,179,1.0);";
```
philip-alldredge, line 20: Please move static constants to top of class for consistency with the project as a whole.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,79 @@
...
+		Button.setAlignment(Pos.BASELINE_LEFT);
+		Button.setGraphic(new ImageView(model.getItemIcon(item)));
+		Button.setOnAction(e -> {
+			model.activateItem(item);
+		});
+		this.getChildren().add(Button);
+		updateStyle();
+	}
+
+	private void resourceChangeListenerHandler(ObservableValue<? extends I> observable, I oldValue, I value) {
```
philip-alldredge, line 50: Remove these print statements. Final code shouldn't contain any print statements like this.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,79 @@
...
+		System.err.println(observable + " observable");
```
philip-alldredge, line 51: Rename this function to something suitable for what it is. It is not a resource change listener. It is a listener for the active item property. Suggest `activeItemChanged`

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,79 @@
...
+import javafx.geometry.Insets;
+import javafx.geometry.Pos;
+import javafx.scene.control.Button;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.Region;
+import javafx.scene.text.Font;
+
+class PaletteItem<I> extends Region {
+
+	private final Button Button;
```
philip-alldredge, line 16: Rename all fields and methods which start with a capital letter. All fields should(With the exception of those used as constants) should have a lower camel case name. Review the standard java naming conventions for other similar conventions.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
```diff
@@ -0,0 +1,79 @@
...
+		Button.setFont(new Font(14));
+		Button.setOnMouseEntered(e -> {
+			Hovered = true;
+			updateStyle();
+		});
+		Button.setOnMouseExited(e -> {
+			Hovered = false;
+			updateStyle();
+		});
+		model.activeItemProperty().addListener(new WeakChangeListener<I>(this::resourceChangeListenerHandler));
```
philip-alldredge, line 39: You need to keep a reference to the inner change listener. (See the note at https://docs.oracle.com/javase/8/javafx/api/javafx/beans/value/WeakChangeListener.html) ,

As it is, the change listener is being garbage collected sometimes and change events are no longer being processed. The inner change listener could be just a field in the class without a separate method. Then just pass that value to the WeakChangeListener constructor. This would ensure that the inner listener exists for as long as the `PaletteItem` exists.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteGroup.java
BrendenMcHan: I have made all the above fields private.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteGroup.java
BrendenMcHan: I have made the above changes.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
BrendenMcHan: I have made the above changes.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
BrendenMcHan: I have removed the print statements.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
BrendenMcHan: I have renamed the function to the suggested name.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
BrendenMcHan: I have fixed the naming convention of all fields and methods.

ge/org.osate.ge.fx/src/org/osate/ge/fx/palette/PaletteItem.java
BrendenMcHan: I have added a reference to the weak change listener.

philip-alldredge: @BrendenMcHan see the comment I just added below. The code does not store a reference to the change listener.

BrendenMcHan:
How would I store a reference to this::activeItemChanged since activeItemChanged is a method?

philip-alldredge:
That syntax is just a shorter form of wrapping it as an object. Look at the documentation for WeakChangeListener's constructor to see the type being used.

BrendenMcHan:
I think I have correctly created a reference to the weak change listener.

philip-alldredge:
Yes, that looks better.

BrendenMcHan:
@philip-alldredge In regards to the appearance of the new collapse button, would you like me to find and implement icons similar, or identical if possible, to the icons used for the current button?

philip-alldredge:
@BrendenMcHan I did some digging and it looks like the current palette manually draws the symbol. There isn't an icon image for it. Please create an SVG icon using inkscape(or other application) that matches it or manually draw the button. I noticed that the button is smaller when it is collapse so that it takes up minimal room. Please replicate that behavior as well.

BrendenMcHan:
@philip-alldredge I think I implemented the correct shrinking behavior of the collapse button when I originally created the button and it's accompanying methods. As for the SVG icon I will begin working on that now.

BrendenMcHan:
@philip-alldredge What folder would I need to place the Icon files in to make them accessible by eclipse so that I can implement them?

philip-alldredge:
Usually we would create an "icons" folder at the root of the plugin project and make sure it is included in the binary build by selecting the folder in the build configuration or the META-INF/MANIFEST.MF. This ensures it's distributed in the built version. Unsure how to reference it from the JavaFX side or if there is a better way of doing it.

BrendenMcHan:
Once it is in the same folder as the file it needs to be included in, I should just be able to reference it directly. Thanks for your help.

BrendenMcHan:
@philip-alldredge I have started working on fixing the warnings and errors in the AgeEditor.java file, and I have noticed that there are 1700 errors and an equal number of warnings. Most of them are from line 0 of the file and contain a message like this: "API analysis aborted for 'name of file' " These errors are all labeled as fatal, and the warnings seem to have the same message as the errors. Would these errors be caused by my code, or are they the results of an improper setup of the environment or possibly another issue?

philip-alldredge:
I imagine it could be from your code if your file isn't compiling. The API analysis is separate from the computer but I imagine that is the code isn't compiling then the API analysis isn't running.  I supposed that is could be caused by an improper development environment configuration but I haven't seen the issue. You can try cleaning. You can also disable the API analysis in this case. There is an item on the eclipse toolbar that looks like a purple circle with an 'I" in it.

BrendenMcHan:
Ok, thank you for the suggestions.

BrendenMcHan:
@philip-alldredge The icon issue should now be fixed

philip-alldredge:
@BrendenMcHan the only "real" error you should be related to org.osate.ge.ba. Just close that project. The other issues are compatibility and version numbering related problems. As discussed the other day, those can be ignored. The currently pushed code does not work.

BrendenMcHan:
I realize the current code doesn't work, the errors I was referring to were the errors that occur when osate is launched and prevent the code from running. I resolved many of those errors yesterday, and I should be able to get the code running today.

philip-alldredge:
@BrendenMcHan . I was commenting because there weren't any errors that preventing launching of OSATE the errors that were occuring were runtime errors.

BrendenMcHan:
I have managed to fix all the runtime errors, but I have noticed that the collapse button icons do not appear, and I am still working on getting the scrollpane to size correctly within each group.

BrendenMcHan:
@philip-alldredge I have pushed a commit that uses more efficient sizing for the scrollpanes for each group, it allows all of them to be opened without taking up too much room and pushing each other too far, this should make things much neater. When you have time, could you review the functionality and make sure its what you expected and let me know if I need to make any changes. Thanks

BrendenMcHan:
@philip-alldredge While working on the Icons for the collapse button, I have come across an issue with the icons not appearing. I have them in the correct folder as you suggested, and upon building, there is no error saying that the URL cannot be recognized. In theory, that means that the project is using the image as the icon, but for one reason or another the icons do not appear where they should. I have tried resizing the icons to see if that was the issue, however that doesn't seem to make a difference. Any advice would be greatly appreciated, thanks.

philip-alldredge:
I did some brief experimentation and it appears to be one or all of the following:
- Size of image.
- Format of image. SVG is good for design and storing the master but it may not be supported by ImageView.
- Button padding.

BrendenMcHan:
I will look into each of those and see if I can get the icons working. Thanks

philip-alldredge:
Closing. Will reopen with other branch when ready.

PR #2255, opened 2020-03-26
1204 ge improve flow tools validation

### LINKED ISSUE #1204: Improved validation in graphical flow creation tools
<a href="https://github.com/philip-alldredge"></img></a> **Issue by [philip-alldredge](https://github.com/philip-alldredge)**
_Wednesday Feb 14, 2018 at 16:38 GMT_
_Originally opened as https://github.com/osate/osate-ge/issues/296_

----

The validation for the flow creation tools has been reduced because they were previously over constrained.  They should be improved to allow the full capabilities of the language but to provide useful validation for when the flow being created would cause an error in the resulting model. Ideally, this would reuse the validation which is part of the text editor.

Additionally, need to test and support creating flows that extend to the subprogram calls. This can be confusing because IIRC, the flow must reference a subprogram subcomponent which is referenced by the call. Should check highlighting in that case too.

RyanMcilnay:
Closes #1204.


RyanMcilnay:
@philip-alldredge If you remember, we discussed validation done by concrete syntax and possibly diagnostician,
example: (this is the order of doing things we discussed)
````
final IProject project = ProjectUtil.getProjectForBoOrThrow(ci);
final ResourceSet testResourceSet = ProjectUtil.getLiveResourceSet(project);
final XtextResource testResource = getXtextResource(testResourceSet, ci.eResource().getURI());
final String serializedSrc = ((XtextResource) ci.eResource()).getSerializer()
					.serialize(ci.eResource().getContents().get(0));
loadResource(testResource, serializedSrc);
final EObject objectToModify = testResourceSet.getEObject(EcoreUtil.getURI(ci), true);
if (resource.validateConcreteSyntax().size() > 0) {
    // "Validation failed"
}

Diagnostic diagnostic = Diagnostician.INSTANCE.validate(objectToModify,
								Collections.singletonMap(Diagnostician.VALIDATE_RECURSIVELY, true));
System.err.println(diagnostic + " diag");

````

But, this does not catch all problems.  I have found that if I do things in this order I can catch the exception when creating flows,
example:
````
....
final ComponentImplementation objectToModify = (ComponentImplementation) testResourceSet
								.getEObject(EcoreUtil.getURI(ci), true);

objectToModify.setNoFlows(false);
objectToModify.getOwnedFlowImplementations().add(flowImpl);

try {
    final String serializedSrc = ((XtextResource) objectToModify.eResource()).getSerializer()
									.serialize(objectToModify.eResource().getContents().get(0));
    loadResource(testResource, serializedSrc);
} catch (final RuntimeException e) {
    System.err.println("Serialize Exception : " + e.getMessage());
}
````

Pretty much catching an exception, and deciding if validation failed or not.  This is not the approach we discussed so I am wondering if this is acceptable.  To be clear, I believe validation would involve all three methods: checking the exception, concrete validation, and the Diagnostician for errors.  But that is just my initial testing, might be able to not use one.

**Comment by philip-alldredge** [comment]
@RyanMcilnay  to be clear, you are saying there are times when serialization fails but the concrete syntax validation passes? If so, we may need to do it that way.

**Comment by RyanMcilnay** [comment]
@philip-alldredge Yep, that is what I am saying.  And okay, I'll go in that direction.  Thanks!

**Comment by philip-alldredge** [comment]
I don't feel like it's the best practice but if we find a better way in the future, we should be able to swap it out.

**Comment by RyanMcilnay** [comment]
@philip-alldredge Got it, that's why I wanted to check with you.  I can spend a little time looking at other options once I get things implemented so I'm not hung up on this issue.

**Comment by philip-alldredge** [comment]
I agree, don't get hung up on it. Thanks!

**Comment by RyanMcilnay** [comment]
@philip-alldredge When you get the chance, can you test creating flow implementations?  (I have not done ETE flows yet).   Keep in mind this is just validation.  I realize more selection restrictions will need to be removed for segment editing (#1212).

**Comment by philip-alldredge** [comment]
@RyanMcilnay  I haven't reviewed it yet, but is this intended to be merged if it all works? If not, please mark it as a draft pull request.

**Comment by RyanMcilnay** [comment]
@philip-alldredge It is not ready to be merged, just wanted you to test the tool.  Also I am not sure how to mark it as a draft.

**Comment by philip-alldredge** [comment]
@RyanMcilnay I should be able to look at it today. it today. Apparently you can't convert a normal pull request to a draft one. You have to select draft from the create button when creating it initially. I've added a WIP prefix to this one. Please remove it when it's ready for final review and merge.

**Comment by RyanMcilnay** [comment]
@philip-alldredge Alright, thanks!  I'll try to remember to select draft when necessary.

**Comment by philip-alldredge** [comment]
There were two branches for #1204. I removed '1204_improve_flow_tools_validation' to avoid confusion.

**Comment by RyanMcilnay** [comment]
@philip-alldredge Okay, got it.

File: ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/CreateFlowImplementationTool.java [hunk 1]
```diff
@@ -157,32 +168,42 @@ public void onSelectionChanged(@Named(Names.BUSINESS_OBJECT_CONTEXTS) final Busi
...
+		CreateFlowImplementationDialog(final Shell parentShell, final UiService uiService,
+				final ColoringService.Coloring coloring) {
 			super(parentShell);
 			this.uiService = Objects.requireNonNull(uiService, "ui service must not be null");
 			this.coloring = Objects.requireNonNull(coloring, "coloring must not be null");
 			this.setHelpAvailable(true);
 			setShellStyle(SWT.CLOSE | SWT.MODELESS | SWT.BORDER | SWT.TITLE | SWT.RESIZE);
 		}
 
+		public void setIsValid(boolean isValid) {
```
philip-alldredge: should be `setValid`

philip-alldredge:
@RyanMcilnay I left a small comment but I didn't do a full review or anything since I assume you were waiting for feedback before cleaning things up. Functionality wise it seems to work well.  The performance isn't as nice as I'd like but there may not be much we can do about that.  Can you look into what part of the validation causes the slow down?

RyanMcilnay:
@philip-alldredge The only slow down I am experiencing seems to be coloring the diagram, I believe.  If you disable coloring does that fix the slow down? 
FYI Starting at around line 590ish should be coloring.clear() and a block of code.  Disable that and see?

philip-alldredge:
Looks like you are correct. I see the problem. I'll open up another issue for the performance issue of the coloring.

RyanMcilnay:
@philip-alldredge Ready for review.

philip-alldredge:
@RyanMcilnay please remove the WIP and update the branch to the latest.

RyanMcilnay:
@philip-alldredge Ready for review

RyanMcilnay:
I'll take a look and fix the issue, but to answer your last question, I wanted to check for errors and warnings because if there are existing warnings, the user will see duplicate messages if the flow they are creating has the same warning as an existing element in the model.

Hopefully that makes sense, we can discuss on a call if you want at some point.  I'll be focusing on your first comments for now.

philip-alldredge:
Okay. So, do warning's prevent the user from creating the ETE flow or flow implementations? We may want to display the warning but not prevent creation.

RyanMcilnay:
Just an overview of the functionality, open for discussion.

While tool is active (open):
Warnings that occur when creating the flow **do not** prevent creation.
Errors that occur when creating the flow **do** prevent creation.

When tool is trying to be launched:
Existing errors and warnings in the model **do** prevent tool from launching.

I wouldn't be opposed to only stopping launch when existing errors (not warnings) are present.  Just made a decision and went with it because I wanted the messages in the table to only pertain to the flow being created.

philip-alldredge:
For now, let's do what you described with regards to warning and errors. We may have to adjust the launching behavior based on user feedback.

RyanMcilnay:
@philip-alldredge I have addressed the feedback from earlier.

philip-alldredge:
@RyanMcilnay 
- When creating end to end flow specifications, there will be errors until the user specifies a name. Some of the errors will not make sense.  I think we should move the name field so it is before the model errors. Perhaps it should be all the way to the top and the user should be prompted to enter the name first. Thoughts?
- Once the UI is finalized, you'll need to update the user guide.
- The error message when there are errors in the model still needs to be updated.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +107,181 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+	public static List<Diagnostic> getAllReferencedPackageDiagnostics(final BusinessObjectContext boc) {
+		// Find root to get all referenced package diagnostics
+		final BusinessObjectContext root = findRoot(boc)
+				.orElseThrow(() -> new RuntimeException("Cannot find root"));
+
+		// Get get root elements
+		final Set<NamedElement> rootElements = root.getAllDescendants().map(queryable -> {
+			final Object bo = queryable.getBusinessObject();
+			return Element.class.isInstance(bo) ? bo : null;
+		}).filter(Predicates.notNull()).map(bo -> ((Element) bo).getElementRoot()).collect(Collectors.toSet());
```
philip-alldredge, line 123: Filter out root elements which are null. Could filter out all those that are not packages.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +107,181 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+	 */
+	public static List<Diagnostic> getAllReferencedPackageDiagnostics(final BusinessObjectContext boc) {
+		// Find root to get all referenced package diagnostics
+		final BusinessObjectContext root = findRoot(boc)
+				.orElseThrow(() -> new RuntimeException("Cannot find root"));
+
+		// Get get root elements
+		final Set<NamedElement> rootElements = root.getAllDescendants().map(queryable -> {
+			final Object bo = queryable.getBusinessObject();
+			return Element.class.isInstance(bo) ? bo : null;
```
philip-alldredge, line 122: I believe `bo instanceof Element` would be cleaner.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/CreateEndToEndFlowSpecificationTool.java
```diff
@@ -81,18 +85,20 @@
...
 			final ColoringService coloringService, final NamingService namingService) {
 		try {
-			ciBoc = ToolUtil.findComponentImplementationBoc(selectedBoc);
-			if (ciBoc != null) {
-				this.ci = (ComponentImplementation) ciBoc.getBusinessObject();
+			// Check for existing errors or warnings
+			final List<Diagnostic> diagnostics = ToolUtil.getAllReferencedPackageDiagnostics(selectedBoc);
+			if (!diagnostics.isEmpty()) {
+				Display.getDefault()
+						.asyncExec(() -> new FlowDialogUtil.ErrorDialog("Cannot create a new end-to-end flow.",
```
philip-alldredge, line 99: Improve error message with something that explains to the user the problem and how to fix it. Example: `Errors detected in model. The Create End-to-End Flow Tool cannot be used until errors have been resolved. Resolve errors and try again.`

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/CreateEndToEndFlowSpecificationTool.java
```diff
@@ -232,29 +246,18 @@ private Element getRefinedElement(final Element ce) {
...
 	private class CreateFlowsToolsDialog extends TitleAreaDialog {
 		private final NamingService namingService;
 		private final UiService uiService;
 		private final Aadl2Package pkg = Aadl2Factory.eINSTANCE.getAadl2Package();
 		private final EndToEndFlow eTEFlow = (EndToEndFlow) pkg.getEFactoryInstance().create(pkg.getEndToEndFlow());
-		private final List<EndToEndFlow> flows = new ArrayList<EndToEndFlow>();
 		private final List<String> segmentList = new ArrayList<String>();
 		private final List<String> modeList = new ArrayList<String>();
+
+		private Optional<EndToEndFlow> flow = Optional.empty();
```
philip-alldredge, line 257: Unlike in other languages, it is usually not recommended to use `Optional` in fields. Replace with a regular field which may be null. ( I don't necessarily agree with this but it is standard java recommendation.)

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/CreateEndToEndFlowSpecificationTool.java
```diff
@@ -267,8 +270,8 @@ public CreateFlowsToolsDialog(final Shell parentShell, final NamingService namin
 			setShellStyle(SWT.CLOSE | SWT.MODELESS | SWT.BORDER | SWT.TITLE | SWT.RESIZE);
 		}
 
-		private List<EndToEndFlow> getFlows() {
-			return Collections.unmodifiableList(flows);
+		private Optional<EndToEndFlow> getFlow() {
```
philip-alldredge, line 273: If you need this method, this is a good use of Optional. You can use Optional.ofNullable() to convert your nullable field into an optional.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/CreateFlowImplementationTool.java
```diff
@@ -80,26 +85,35 @@
...
-			update(new BusinessObjectContext[] { selectedBoc }, true);
-			if (dlg.open() == Window.CANCEL) {
-				return;
-			}
+			// Check for existing errors or warnings
+			final List<Diagnostic> diagnostics = ToolUtil.getAllReferencedPackageDiagnostics(selectedBoc);
+			if (!diagnostics.isEmpty()) {
+				Display.getDefault()
+				.asyncExec(
+						() -> new FlowDialogUtil.ErrorDialog("Cannot create a new flow implementation.",
```
philip-alldredge, line 96: See comment on ETEF tool.

RyanMcilnay:
How do you feel about a default name like "new_ete_flow" being set when the dialog is launched?  That would get rid of the weird errors while creating.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
RyanMcilnay: Alright, I'll refrain from using it in the future as well.

philip-alldredge:
> @philip-alldredge How do you feel about a default name like "new_ete_flow" being set when the dialog is launched? That would get rid of the weird errors while creating.

I'm find with that as long as the naming server or similar is used to generate a unique name. We don't want the default name to conflict.

RyanMcilnay:
@philip-alldredge I addressed the comments and did a little cleanup.  Ready for review again.  I'll work on the documentation and get some screenshots once the UI is approved, like you mentioned.

RyanMcilnay:
What are you selecting first that you are getting unrecognizable errors?

philip-alldredge:
What do you mean "unrecognizeable errors"? Is my last review, for the naming issues, I was selecting a flow source when the names were invalid. Other selections resulted in the name being blank.

RyanMcilnay:
You have a "causes same issue as previously discussed" and I assumed the wrong issue because we have discussed a lot of issues.  I thought you were referring to "When creating end to end flow specifications, there will be errors until the user specifies a name. Some of the errors will not make sense..."

RyanMcilnay:
Really weird, my text field is not 1 character behind either.

philip-alldredge:
What I was referring to is that the cause is the same as in you need to know what component implementation isn't being modified to check the resource for existing errors when opening the dialogs.

RyanMcilnay:
Never mind, text field and errors are behind.

RyanMcilnay:
Philip, our discussion earlier mentioned showing the user the qualified name of the component implementation being modified, and we decided to hold off on that.  I thought about it and it is probably a good idea and was not hard to add to the interface.  So that is added now.

Code is ready for review and testing once again.  Thanks.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +108,234 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+	private static void loadResource(final XtextResource resource, final String src) {
+		try {
+			resource.unload();
+			resource.load(new ByteArrayInputStream(src.getBytes()), null);
+		} catch (final IOException e) {
+			throw new RuntimeException("Serialized source cannot be loaded");
+		}
+	}
+
+	private static XtextResource getXtextResource(final ResourceSet resourceSet, final URI uri) {
```
philip-alldredge, line 335: I'd recommend renaming this to `getOrCreateXtextResource`

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +108,234 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+			return (this.message.equalsIgnoreCase(e.getMessage()));
+		}
+
+		@Override
+		public int hashCode() {
+			return this.message.hashCode();
+		}
+	}
+
+	private static boolean isErrorOrWarning(final int severity) {
```
philip-alldredge, line 292: It would be helpful to add a javadoc indicating what `severity` is since it is just an int. It's not clear without looking into the code.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +108,234 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+			return new DiagnosticWrapper(severity,
+					getDetailedDiagnosticMessage(this.msgPrefix, diagnostic.getMessage()));
+		}
+
+		private static String getDetailedDiagnosticMessage(final String prefix, final String message) {
+			return new StringBuilder(prefix).append(message).toString();
+		}
+	}
+
+	private static class DiagnosticWrapper extends BasicDiagnostic {
```
philip-alldredge, line 264: Add a javadoc comment describing the purpose of this class. Rename if there is a better name. It is not clear to me why this exists.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
```diff
@@ -80,4 +108,234 @@ public static Context findContextExcludeOwner(final BusinessObjectContext boc,
...
+			if (getClass() != o.getClass()) {
+				return false;
+			}
+
+			final DiagnosticWrapper e = (DiagnosticWrapper) o;
+			return (this.message.equalsIgnoreCase(e.getMessage()));
+		}
+
+		@Override
+		public int hashCode() {
```
philip-alldredge, line 287: Why are hashCode and equals being overriden? This implementation violates the contract with `hashCode` Equal object must return the same hashCode.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
RyanMcilnay: This exists so duplicate error messages do not return.  For instance, when name conflicts occur, two errors are returned.  One error for the flow being created and one for the flow that exists.  This is this way the validation works, but it did seem a little redundant.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
philip-alldredge: One way to make it consistent is to store a lowercase version of the message as  a private field and to use it in hashcode and equals.

philip-alldredge:
@RyanMcilnay  the build failed because of the version used for the coverage report. Please see https://github.com/osate/osate2/wiki/Managing-Feature-and-Plugin-Versions#building-orgosatebuildcoveragereport-fails-with-tychomaven

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
RyanMcilnay: I made changes based on the first comment, probably shouldn't be ignoring severity anyways.

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
RyanMcilnay: I changed this to make sense and added javadoc

ge/org.osate.ge/src/org/osate/ge/internal/ui/tools/ToolUtil.java
philip-alldredge: Well, I don't think my original comment was clear. The original problem wasn't the severity, it was that hashCode was that equals was doing a case insensitive comparison and that wasn't consistent with hashCode. Although I agree that severity shouldn't be ignored.

RyanMcilnay: Yeah, I know, my updated code fixes both issues right?

philip-alldredge: Correct but I wanted to be clear on the original issue.

That said, I'm not sure if all the changes are beneficial. The new name `BasicDiagnostic` is confusing since EMF already has one and the new javadoc isn't very helpful. All diagnostics contain severity and a message. Based on your earlier comments, the purpose of this diagnostic implementation is to avoid duplicates. That's the important part and it is not captured in the comments.

However, I think there may be a better way of handling things. I realize this may require changes in several places, but what do you think about using TreeSet with a custom comparator instead of HashSet. The comparator would compare the diagnostics in whatever way is appropriate. Just comparing may be okay.. If you did this, you wouldn't need the new diagnostic type. The builder could just create EMF BasicDiagnostic directly.  Thoughts? 

In any case, let me know when you finish the user guide updates and I'll do a final review.

RyanMcilnay: Well, XtextResource returns a org.eclipse.emf.ecore.resource.Resource.Diagnostic

philip-alldredge: I'm not sure if I see the issue.

Diagnostic isn't a class, it is an interface. Ideally we would write everything to use Diagnostic. You are adding a prefix so you subclassed EMF BasicDiagnostic. Correct? However, if you use TreeSet with a comparator, you can continue to use EMF's `BasicDiagnostic` without our separate class.  Whatever route you take, we need a comment/javadoc explaining why we are doing it that way.

philip-alldredge:
@RyanMcilnay I pushed a change because of an issue with `toDiagnosticTreeSet`. The comparator should return 0 for equal values. Otherwise, I believe there is a chance odd things can occur. To be safe, I went ahead and made the change. Please make a quick review and let me know if it looks right to you.

RyanMcilnay:
@RyanMcilnay looks fine to me

PR #2286, opened 2020-04-16
Updated Instantiate action to work on AADL files

### LINKED ISSUE #2272: Instantiate action should work on .aadl files in the AADL Navigator
Currently to instantiate a model you need to either

1. Open an `.aadl` file, then select the component instances in the outline view, and then choose Instantiate from the context menu, or

2. Drill into the `.aadl` file in the navigator and then choose instantiate from the context menu.

Number 1 is annoying because I don't want to open the declarative model file in an editor just to instantiate.

Number 2 is annoying because I there are 2 levels to unwrap in the navigator before the system classifier nodes are uncovered.

I propose that when the instantiate command is executed on an declarative model, a dialog appears with the available component classifiers for the user to select from and instantiate.  

Furthermore, I think that there should be a setting the controls the behavior: by default all system implementation classifiers in the declarative file are instantiated without showing a dialog box.

AaronGreenhouse:
Fixes #2272


core/org.osate.core/src/org/osate/core/OsateCorePlugin.java
```diff
@@ -339,4 +349,32 @@ public final int getSOMLimit() {
...
+	}
+
+	public static void setOnlyInstantiateSystemImplsDefault(final IPreferenceStore store) {
+		store.setDefault(OsateCorePlugin.ONLY_INSTANTIATE_SYSTEM_IMPLS,
+				OsateCorePlugin.ONLY_INSTANTIATE_SYSTEM_IMPLS_DEFAULT);
+	}
+
+	public static String getOnlyInstantiateSystemImplsPreferenceName() {
+		return ONLY_INSTANTIATE_SYSTEM_IMPLS;
+	}
```
lwrage, line 379: remove, use constant, instead

PR #2301, opened 2020-04-28
Added flow spec name to the component name in the latency output

### LINKED ISSUE #2289: Flow latency report should include flow names in leaf components
## Summary

The report created by the flow latency analysis include the names of leaf components but not of the flow, which makes it hard to follow if there are multiple flows.

The flow source/path/sink name should be included in the report.

## Environment

* **OSATE Version**: 2.7.0
* **Operating System**: Linux

AaronGreenhouse:
Fixes #2289


lwrage:
It's strange that you had to update a BA plug-in version. It shouldn't be necessary.

AaronGreenhouse:
Updated the docs to include the changes to the analysis results object structure.

PR #2305, opened 2020-04-30
fix highlighting of refined flow segments

### LINKED ISSUE #2304: Refined flow segments are not highlighted
Flow segments of End To End Flows and Flow Implementations that are refined are not highlighted in graphical editor.

RyanMcilnay:
closes #2304 .


philip-alldredge:
@RyanMcilnay  Please see the attached project. Highlighting of either of the flow implementation does not work as expected. This does not even account for the flows themselves being refined. It illustrates the possibility that the flow refers to a refined element instead of the base. You probably want to compare the "root" refined element on both sides of the comparisons.


[2304_flow_refinement.zip](https://github.com/osate/osate2/files/4559258/2304_flow_refinement.zip)

philip-alldredge:
@RyanMcilnay is this ready for another review?

RyanMcilnay:
@philip-alldredge Yep, it is.

PR #2316, opened 2020-05-05
2306 ge properties for refined elements

### LINKED ISSUE #2306: Graphical editor AADL properties section of refined elements doesn't work
Refined element properties not checking refined element recursively.

RyanMcilnay:
Closes #2306.


philip-alldredge:
@RyanMcilnay can you explain what this fixes? The issue isn't very clear about what the problem is.

ge/org.osate.ge/src/org/osate/ge/internal/ui/properties/ChangeSubcomponentTypePropertySection.java
```diff
@@ -140,11 +141,7 @@ public void refresh() {
 	}
 
 	private static boolean isCompatibleSupcomponentType(final NamedElement ne, final EClass subcomponentType) {
```
philip-alldredge, line 143: Can you fix the spelling of subcomponent since you are modifying this function?

ge/org.osate.ge/src/org/osate/ge/internal/ui/properties/SwitchDirectionOfConnectionPropertySection.java
```diff
@@ -102,11 +103,7 @@ public void widgetSelected(final SelectionEvent e) {
 	};
 
 	private static Connection getEditConnection(final Connection connection) {
-		if (connection.getRefined() != null) {
-			return connection.getRefined();
-		}
-
-		return connection;
+		return (Connection) AadlHelper.getRootRefinedElement(connection);
```
philip-alldredge, line 106: This part makes sense. No need to explain.

RyanMcilnay:
@philip-alldredge Fixed based on your comments.

ge/org.osate.ge/src/org/osate/ge/internal/ui/properties/ChangeSubcomponentTypePropertySection.java
```diff
@@ -140,11 +141,7 @@ public void refresh() {
...
-		Subcomponent sc = (Subcomponent) ne;
-		if (sc.getRefined() != null) {
-			sc = sc.getRefined();
-		}
-
+		final Subcomponent sc = (Subcomponent) AadlHelper.getRootRefinedElement(ne);
```
philip-alldredge, line 144: I'll try to ignore problems like the fact `getTypeOptionsInformation` does something different from its name IMHO, which are outside the scope of this issue, I don't believe this fix is what we want.

I'm having trouble understanding why the original code gets the refined element at the start. It would seem that the  if statement should be removed.  The compatible subcomponent should be based on the containing classifier of the subcomponent(even if it is refined) and if a AbstractSubcomponent is being refined, then it should be changeable to anything. 

In short, please avoid getting the refined subcomponent at the start and test.

philip-alldredge: @RyanMcilnay sorry I added the last comment after you made changes. I should of taken my own advise and used the review feature.

RyanMcilnay:
@philip-alldredge Changes made based on comments are done.

PR #2330, opened 2020-05-14
Adjusts labels for graphical editor flow combo. 

### LINKED ISSUE #2328: Improve performance with the flow contribution item
When working with models with many packages and flows, the flow contribution item calls `getSegmentState` on each flow to determine what suffix to add to the label. This results in a significant load time and opens related resources.

Need to investigate and optimize. Is it loading more than necessary? It is likely working as intended. However, it might be best to avoid eagerly calculating and showing the suffixes. The information is mostly reflected in the state of the `Show` button. If limited to that, it should be practical to only calculate it for the state selected item.

philip-alldredge:
Closes #2328. Lazily compute state as needed. Removed prefix for flows. This greatly improves performance for larger models.


RyanMcilnay:
@philip-alldredge I haven't looked into it, but my initial thought is:
Do we need all three states empty, partial, and complete still?  It would seem like we no longer care if flows are empty, we would be satisfied with a complete or missing state.  There may be some code that is checking for empty flows that could be shortcut with the first missing segment.

What are your thoughts?

Maybe that is something we can cleanup later, this does seem to fix the current issue.

philip-alldredge:
You are correct there may be a slight improvement for removing the distinction between empty and partial. I'd prefer that we wait until later to do that. We may decide we want to distinguish them at some point after we've given this more though. The current pull request fixes the main performance issue and doesn't change too much code.

PR #2338, opened 2020-05-18
Spotbugs for org.osate.analysis.resource

AaronGreenhouse:
Fixed some issues, I hope.


analyses/org.osate.analysis.resource.management/src/org/osate/analysis/scheduling/TestDataInputStream.java
```diff
@@ -36,10 +36,14 @@
 
 public class TestDataInputStream {
```
lwrage, line 37: I would probably better to delete this whole class.

AaronGreenhouse:
Done with this.  FIxed the .analysis errors except for the binpacker

PR #2358, opened 2020-06-08
More SonarCloud issues

keh181:
Here are some more SonarCloud fixes. The commit for BinPackerTester asked for me to increment the version but only 2 files needed updated.


keh181:
BinPackerTester commit causes the Jenkins test to fail because an API was incremented

PR #2367, opened 2020-06-17
SonarCloud bugs and vulnerabilities

keh181:
More fixes for Sonar Cloud issues.


keh181:
Requires	the following versions to be increased:
- /org.osate.utils/META-INF/MANIFEST.MF
-  /org.osate.importer.simulink/META-INF/MANIFEST.MF
-  /org.osate.annexsupport/META-INF/MANIFEST.MF
- /org.osate.analysis.binpacking/META-INF/MANIFEST.MF

lwrage:
@joeseibel Could you review this, please?

ba/org.osate.utils/src/org/osate/utils/FileUtils.java
```diff
@@ -140,12 +140,13 @@ public static void copyFiles(List<File> srcFiles, File destFolder, List<String>
...
-		out = new BufferedWriter(new FileWriter(inFile));
-		out.write(string2write);
-		out.newLine();
-		out.close();
+	public static void copyIntoFile(File inFile, String string2write) {
+		try (BufferedWriter out = new BufferedWriter(new FileWriter(inFile))) {
+			out.write(string2write);
+			out.newLine();
+		} catch (IOException e) {
+			e.printStackTrace();
```
joeseibel, line 148: The original method method propagated any `IOException`s. This version should do the same. Keep the try-with-resources, but don't catch the `IOException` and let it be propagated to the caller.

core/org.osate.xtext.aadl2.properties/src/org/osate/xtext/aadl2/properties/util/GetProperties.java
```diff
@@ -864,11 +864,11 @@ public static double getMinimumTimeToTransferData(final NamedElement bus, Classi
 	}
 
 	public static double fromMStoSec(NamedElement ne, double value) {
-		return convertToScale(value, getMSUnitLiteral(ne), getSecUnitLiteral(ne));
+		return getMSUnitLiteral(ne) != null ? convertToScale(value, getMSUnitLiteral(ne), getSecUnitLiteral(ne)) : -1;
```
joeseibel, line 867: I'm not sure that returning -1 is the right thing to do here. Under normal conditions, I would expect that `getMSUnitLiteral()` would never return `null`. If `ms` cannot be found, then something very serious has gone wrong. It is probably better to throw a `NullPointerException` which complains that the unit literal could not be found.

core/org.osate.xtext.aadl2.properties/src/org/osate/xtext/aadl2/properties/util/GetProperties.java
```diff
@@ -864,11 +864,11 @@ public static double getMinimumTimeToTransferData(final NamedElement bus, Classi
...
 	}
 
 	public static double fromUStoSec(NamedElement ne, double value) {
-		return convertToScale(value, getUSUnitLiteral(ne), getSecUnitLiteral(ne));
+		return getUSUnitLiteral(ne) != null ? convertToScale(value, getUSUnitLiteral(ne), getSecUnitLiteral(ne)) : -1;
```
joeseibel, line 871: I'm not sure that returning -1 is the right thing to do here. Under normal conditions, I would expect that `getUSUnitLiteral()` would never return `null`. If `us` cannot be found, then something very serious has gone wrong. It is probably better to throw a `NullPointerException` which complains that the unit literal could not be found.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -3145,7 +3145,7 @@ private void checkComponentTypeModes(ComponentType componentType) {
 		} else if (componentType.isDerivedModes()) {
 			INode requiresModesNode = getRequiresModesNode(componentType);
 			if (componentType.getExtended() != null && componentType.getExtended().getAllModes().stream()
-					.anyMatch(extendedMode -> !extendedMode.isDerived())) {
+					.anyMatch(extendedMode -> !extendedMode.isDerived()) && requiresModesNode != null) {
```
joeseibel, line 3148: It is a bit unfortunate that SonarCloud is complaining about a potential NPE for `requiresModesNode`. It is guaranteed to not be null because `componentType.isDerivedModes()` is true, but SonarCloud has no way of knowing that. I don't like the idea of including a null check in the conditional after `anyMatch` because I think it confuses the logic of the validator. When reading validation code and trying to understand what it is doing, the if-statements are critical to figuring out what the potential cases are that the validation method can deal with. I don't like the idea of mixing legitimate conditionals with a conditional that will always evaluate to true and is only there because of SonarCloud.

I would suggest moving the declaration of `requiresModesNode` down into the if body and then doing a separate null check before the call to `acceptError`. Also, put in a comment stating that `requiresModesNode` should never be null and that this check is only there to satisfy SonarCloud.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -3217,7 +3217,7 @@ private void checkExtensionAndRealizationHierarchy(ImplementationExtension imple
 		}
 		if (!isAncestor) {
 			error(implementationExtension, '\'' + typeOfParent.getQualifiedName() + "' is not an ancestor of '"
-					+ typeOfChild.getQualifiedName() + "'.");
+					+ typeOfChild != null ? typeOfChild.getQualifiedName() : "TYPE OF CHILD NOT FOUND" + "'.");
```
joeseibel, line 3220: A null check is already performed on `typeOfChild` a few lines earlier with a call to `Aadl2Util.isNull(typeOfChild)`. I think we should ignore this issue reported by SonarCloud.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -3425,7 +3425,8 @@ private void checkFeaturesOfExtendedAbstractType(ComponentType componentType) {
 		for (FeatureType featureType : typesOfInheritedFeatures) {
 			if (!acceptableFeatureTypes.contains(featureType)) {
 
-				String changeFrom = typeCategory.getName();
+				String changeFrom = typeCategory != null ? typeCategory.getName()
```
joeseibel, line 3428: I think this can be resolved in a better way. A few lines earlier is a bunch of if-else-ifs that perform `instanceof` checks and assign values to `typeCategory`. The if-else-ifs should be exhaustive. I think it would be better to add a last else block and throw an `AssertionError` complaining that there was an unrecognized `ComponentType`.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -3487,7 +3488,8 @@ private void checkSubcomponentsOfExtendedAbstractImplementation(ComponentImpleme
 		for (ComponentCategory subcomponentCategory : categoriesOfInheritedSubcomponents) {
 			if (!acceptableSubcomponentCategories.contains(subcomponentCategory)) {
 				error(componentImplementation.getOwnedExtension(),
-						"A " + implementationCategory.getName()
+						"A " + implementationCategory != null ? implementationCategory.getName()
```
joeseibel, line 3491: I think this can be resolved in a better way. A few lines earlier is a bunch of if-else-ifs that perform `instanceof` checks and assign values to `implementationCategory`. The if-else-ifs should be exhaustive. I think it would be better to add a last else block and throw an `AssertionError` complaining that there was an unrecognized `ComponentImplementation`.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -6259,7 +6261,7 @@ private void checkAccessConnectionProvidesRequires(AccessConnection connection)
 
 		// Test for L5: connection between access features of sibling components
 		if (srcContext instanceof Subcomponent && dstContext instanceof Subcomponent && source instanceof Access
-				&& destination instanceof Access) {
+				&& destination instanceof Access && destinationType != null) {
```
joeseibel, line 6264: There is a better way of doing this instead of a null check. Since `AccessType` is an enumeration, we can use `==` instead of `.equals()` to compare the enum literals. Use `==` for the comparisons of `sourceType` and `destinationType` to `AccessType.PROVIDES`.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -6283,7 +6285,7 @@ else if (source instanceof Access && destination instanceof Access
 		}
 		// Test for L6: connection between subcomponent and access feature
 		else if (source instanceof Subcomponent && destination instanceof Access
-				&& (dstContext == null || dstContext instanceof FeatureGroup)) {
+				&& (dstContext == null || dstContext instanceof FeatureGroup) && destinationType != null) {
```
joeseibel, line 6288: There is a better way of doing this instead of a null check. Since `AccessType` is an enumeration, we can use `==` instead of `.equals()` to compare the enum literals. Use `==` for the comparison of `destinationType` to `AccessType.PROVIDES`.

core/org.osate.xtext.aadl2/src/org/osate/xtext/aadl2/validation/Aadl2JavaValidator.java
```diff
@@ -6302,7 +6304,7 @@ else if (destination instanceof Subcomponent && source instanceof Access
 		// Test for L7: connection between subcomponent and access feature of
 		// subcomponent
 		else if (source instanceof Subcomponent && destination instanceof Access
-				&& dstContext instanceof Subcomponent) {
+				&& dstContext instanceof Subcomponent && destinationType != null) {
```
joeseibel, line 6307: There is a better way of doing this instead of a null check. Since `AccessType` is an enumeration, we can use `==` instead of `.equals()` to compare the enum literals. Use `==` for the comparison of `destinationType` to `AccessType.REQUIRES`.

analyses/org.osate.analysis.binpacking/src/EAnalysis/BinPacking/MacroPreprocessor.java
```diff
@@ -84,10 +85,10 @@ public void loadMacros(File file) {
 	}
 
 	public void substituteMacrosInFile(File file) {
-		try {
+		PrintWriter writer = null;
+		try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
 			File temp = File.createTempFile("TMP", "MPP");
-			PrintWriter writer = new PrintWriter(new FileOutputStream(temp));
-			BufferedReader reader = new BufferedReader(new FileReader(file));
+			writer = new PrintWriter(new FileOutputStream(temp));
```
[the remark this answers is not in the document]
keh181: I originally didn't use try-with-resources for `PrintWriter` because of the variable `temp`, and I still can't seem to be able to come up with a way to use it. Do you have any suggestions on how I could do this? Or should I instead close the `BufferedReader` in a the `finally` block so that they're handled the same way?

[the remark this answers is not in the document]
joeseibel: You can do this with two `try` blocks. The whole thing is wrapped in a `try` ... `catch (IOException)` and then have an inner try-with-resources:

```
try {
	File temp = File.createTempFile("TMP", "MPP");
	try (PrintWriter writer = new PrintWriter(new FileOutputStream(temp));
			BufferedReader reader = new BufferedReader(new FileReader(file))) {
		// Do stuff with writer and reader.
	}
	String fname = file.getAbsolutePath();
	file.renameTo(new File("tmp.old"));
	temp.renameTo(new File(fname));
} catch (IOException e) {
	e.printStackTrace();
}
```

PR #2379, opened 2020-07-08
Graphical editor API rework and refactoring

### LINKED ISSUE #2336: Refactor, cleanup, and rework API for graphical editor
This issue has been reworked. The original issue was too broad and will require to great of a time commitment. We will not split the current code into several plugins at this time. There may be limited renaming and shuffling files to new packages. Moving forward, we want to structure the code base to allow splitting code into plugins later.

- [x] Rework API. Remove DI and review.
  - [x] Palette Items
  - [x] Business Object Providers
  - [x] Tooltips
  - [x] Business Object Handlers
  - [x] References
  - [x] Tools
  - [x] Review and fix API warnings as appropriate
- [x] Reduce Graphiti dependencies

philip-alldredge:
Closes #2336


RyanMcilnay:
Some initial problems I am having, as far as testing goes:

- Creating a feature connection causes a read/write transaction exception, not sure if this is due to my model (I wouldn't think so, but I can send my model if necessary).  

- Creating Subprogram Call Sequences causes a "Unable to determine business object to modify" exception, I think the can create method must be wrong.

- Creating an Access Connection from a subcomponent within another subcomponent, to a feature that is in the containing impl throws an exception, but should probably not be allowed in the first place.  To clarify (hopefully), in my test, I have an impl that contains a subcomponent, and that subcomponent contains another subcomponent.  I am creating a connection that is from the impl, to the subcomponent that is two levels deep.

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorBehaviorStateHandler.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Override
+	public boolean isApplicable(final IsApplicableContext ctx) {
+		return ctx.getBusinessObject(ErrorBehaviorState.class).map(bo -> bo.getElementRoot() instanceof AadlPackage)
+				.isPresent();
+	}
+
+	@Override
+	public CanonicalBusinessObjectReference getCanonicalReference(final ReferenceContext ctx) {
+		final ErrorBehaviorState typedBo = ctx.getBusinessObject(ErrorBehaviorState.class).get();
```
RyanMcilnay, line 59: Possible handling of exception better, instead of calling get(). 
Could be something like:
````
ctx.getBusinessObject(ErrorBehaviorStateMachine.class).map(sm -> new CanonicalBusinessObjectReference(ErrorModelReferenceUtil.TYPE_BEHAVIOR_STATE_MACHINE,
ctx.getReferenceBuilder().getCanonicalReference(sm.getElementRoot()).encode(), sm.getName()))
.orElseThrow(() -> new RuntimeException("Cannot find canonical reference"));
````

Note: This is being done in quite a few places.  Maybe throw an exception in the getBusinessObject method instead of changing code everywhere.  I would like your input on this as well.

File: ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorBehaviorStateHandler.java [hunk 2]
```diff
@@ -0,0 +1,100 @@
...
+		return new CanonicalBusinessObjectReference(ErrorModelReferenceUtil.TYPE_BEHAVIOR_STATE,
+				ctx.getReferenceBuilder().getCanonicalReference(typedBo.eContainer()).encode(), typedBo.getName());
+	}
+
+	@Override
+	public RelativeBusinessObjectReference getRelativeReference(final ReferenceContext ctx) {
+		return new RelativeBusinessObjectReference(ErrorModelReferenceUtil.TYPE_BEHAVIOR_STATE,
+				ctx.getBusinessObject(ErrorBehaviorState.class).get().getName());
```
RyanMcilnay, line 67: See previous comment about calling get()

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorBehaviorStateHandler.java
philip-alldredge: An exception needs to thrown if a canonical reference cannot be retrieved. We are assuming that the business object being provided is of the desired type so we are using `get()`. This is a safe assumption because `isApplicable` checks the type and other methods aren't called unless the handler is applicable. Usually it is good idea to avoid `get` but I believe this is one case where it is reasonable to use. If/when OSATE uses Java 11, we'll start using `orElseThrow()` (without arguments) for cases like this. That method is only in Java 11 and was added because they realized that they shouldn't have named it `get`.

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorBehaviorStateHandler.java
philip-alldredge: Same

philip-alldredge:
@RyanMcilnay once this is approved and merged, then we can fine things that slipped through the cracks during the flow editing as part of #2392.

Issue #2381, opened 2020-07-10
Assurance View does not correctly render subclaims


dvanbrug:
## Summary
When creating a Verification Plan with nested subclaims, the “Assurance View” does not correctly render the Evidence or Description fields for the subclaims.  All of the subclaims inherit the parent’s name / description instead of their own. 

## Expected and Current Behavior
Define a reqspec top level claim C1 and two nested claims C2 and C3 in the following structure:

* C1 – Claim 1     
  * C2 – Claim 2     
  * C3 – Claim 3

I would expect the “Assurance View” to display the Assurance Plan in a similar structure.  Instead, it outputs the following:

* C1 – Claim 1     
  * C1 – Claim 1     
  * C1 – Claim 1

## Steps to Reproduce
1.) Start with the following Alisa example: SimpleControlSystem (https://github.com/osate/alisa-examples/blob/master/SimpleControlSystem)

  * An example of nested subclaims can be seen in the SimpleControlSystem/dualscsvplan.verify file:
https://github.com/osate/alisa-examples/blob/master/SimpleControlSystem/SimpleControlSystem/alisa/dualscsvplan.verify

2.) However, this verification plan is not used in an Assurance Case.  If we modify the SCSCase (SCSVerification.alisa) by adding the following assurance plan entry, we can cause these to be evaluated.

```
 assurance plan DualSCSPlan for SimpleControlSystem::SCSDualPower.impl [
		assure DualSCSvplan
	]
```

3.) Add implementation to SimpleControlSystem.aadl

```
 system implementation SCSDualPower.impl
end SCSDualPower.impl; 
```

4.) Open the "Assurance View" and select the SCSCase Assurance Case.  You should see something similar to the following image:

![OSATE_bugfix](https://user-images.githubusercontent.com/1202619/87155192-83c09a80-c288-11ea-9a65-c2bb08e7d409.PNG)

However, instead it should look the the next message:

![OSATE_bugfix_2](https://user-images.githubusercontent.com/1202619/87155238-99ce5b00-c288-11ea-841b-3466072786e5.PNG)

## Environment

* **OSATE Version**: 2.7.1
* **Operating System**: All

## Proposed Fix
The following method `constructClaimReferencePath` is responsible for building the NestedClaimReference object that is used by the Assurance View.  It appears that the method never adds the subclaim object.

[...]

PR #2382, opened 2020-07-10
Fix NestedClaimReference construction

dvanbrug:
Per #2381 , Subclaims are not correctly rendered in the ALISA Assurance View.

The following method `constructClaimReferencePath` is responsible for building the NestedClaimReference object that is used by the Assurance View.  It appears that the method never adds the subclaim object.

Lines 490-498 in AssureConstructor.xtend https://github.com/osate/osate2/blob/f3d0613e0869dc8830e6ff0f5e43fc9f79e3b185/alisa/org.osate.assure/src/org/osate/assure/generator/AssureConstructor.xtend#L490-L498

If we add the following line after the creation of parentNcr (line 495), the subclaims are correctly added to the parent object.

``` parentNcr.sub = ncr ```


dvanbrug:
@joeseibel Sorry, I did not mean to re-request a review.  I was just taking a look at the status.

PR #2378, opened 2020-07-07
1212 ge flow editing

RyanMcilnay:
Closes #1212

PR #2393, opened 2020-07-20
Fixed version errors

joeseibel:
I think that #2378 missed a few version increments. @philip-alldredge or @RyanMcilnay, could you please review and see if #2378 did indeed miss these version increments?


joeseibel:
The specific errors I was encountering seem to have been fixed by #2379.

PR #2426, opened 2020-09-01
Updated BusLoad analysis to instantiate on run

### LINKED ISSUE #841: Run analysis on implementation
For analyses that operate on the instance model, it should be possible to run them on an implementation. The analysis should automatically generate the instance before running the analysis. This request came up during the AADL standards committee meeting.

AaronGreenhouse:
Fixed #841 -- Analysis can be be launched on a system implementation and the instance will be created and then analyzed.


core/org.osate.ui/plugin.xml
```diff
@@ -525,6 +525,259 @@ censes only apply to the Third Party Software and not any other portion of this
...
+          <and>
+             <instanceof value="org.eclipse.core.resources.IFolder" />
+             <test
+                   property="org.eclipse.core.resources.projectNature"
+                   value="org.osate.core.aadlnature" />
+             </and>
+       </definition>
+
+       <!-- Instance model file -->
+       <definition id="org.osate.ui.definition.private.isAAXLFile">
```
lwrage, line 555: better name: isInstanceFile

core/org.osate.ui/plugin.xml
```diff
@@ -525,6 +525,259 @@ censes only apply to the Third Party Software and not any other portion of this
...
+                type="org.eclipse.core.resources.IFile">
+             <test
+                   property="org.eclipse.core.resources.contentTypeId"
+                   value="org.osate.core.instanceModelFile">
+             </test>
+          </adapt>
+       </definition>
+
+       <!-- Declarative model file: AADL Package -->
+       <definition id="org.osate.ui.definition.private.isAADLFile">
```
lwrage, line 566: better name: isPackageFile
add a definition isPropertySetFile

core/org.osate.ui/plugin.xml
```diff
@@ -525,6 +525,259 @@ censes only apply to the Third Party Software and not any other portion of this
              </adapt>
           </or>
        </definition>
+
```
lwrage, line 528: Shouldn't the simpler definitions be public because they have higher reuse potential?
Should we even distinguish between private and public definitions?

core/org.osate.ui/plugin.xml
AaronGreenhouse: Renamed

core/org.osate.ui/plugin.xml
AaronGreenhouse: Renamed and added

core/org.osate.ui/plugin.xml
AaronGreenhouse: Got rid of `.public` and `.private`

lwrage (approved):

PR #2431, opened 2020-09-03
2429 classifier view npe

### LINKED ISSUE #2429: NPE in classifier information view
## Summary

If a named element (e.g., a flow implementation) has no name the classifier information view throws an NPE and does not display any classifier contents in the right column. An NPE also occurs for unnamed mode transitions.

Missing names should be handled:

- If the name is optional a null name should be interpreted as an empty string (mode transitions)
- Use the flow specification name for a flow implementation

## Steps to Reproduce

1. Load model below into editor
2. Select S.i in outline view 
3. Right click > Open in classifier information view
4. View is empty
5. Similar for A

```aadl
package Issue2429
public
	
	system S
		features
			i: in feature;
			o: out feature;
		flows
			fp: flow path i -> o;
	end S;
	
	system implementation S.i
		flows
			-- NPE when sorting >1 flow implementations
			fp: flow path i -> o;
			fp: flow path i -> o;
	end S.i;
	
	abstract A
		features
			e: in event port;
		modes
			-- NPE when sorting modes and transitions
			-- with unnamed transition
			m1: initial mode;
			m2: mode;
			m1 -[e]-> m2;
	end A;
	
end Issue2429;
```

## Environment

* **OSATE Version**: 2.8.0
* **Operating System**: Win 10, Linux

lwrage:
fixes #2429

### DIFF
--- a/core/org.osate.aadl2/src/org/osate/aadl2/impl/FlowImplementationImpl.java
+++ b/core/org.osate.aadl2/src/org/osate/aadl2/impl/FlowImplementationImpl.java
@@ -139,6 +139,7 @@ public class FlowImplementationImpl extends ModalPathImpl implements FlowImpleme
 	 * <!-- end-user-doc -->
 	 * @generated
 	 */
+	@Override
 	protected EClass eStaticClass() {
 		return Aadl2Package.eINSTANCE.getFlowImplementation();
 	}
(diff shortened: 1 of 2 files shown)


AaronGreenhouse (approved):
Changes are minimal and reasonable.

PR #2453, opened 2020-09-22
improve experimental EMV graphical support

### LINKED ISSUE #1546: Experimental visualization and editing capability for EMV2 flow propagations
**This issue has been revised multiple times to more accurately describe the changes being made**

- [x] Type Library:
	- [x] Rework current implementation. Remove GE dummy model object. No separate graphical element. Types will be contained directly in packages.
	- [x] Palette commands. Automatically add imports as needed.
		- [x] Error type.
		- [x] Error type alias. Prompt on create.			
		- [x] Error type set. Can't be empty. Prompt for type contents when settings.
		- [x] Error type set alias. Prompt on create.
	- [x] Basic diagram support
	- [x] Property section. 
		- [x] Allow setting extended type libraries.
		- [x] Set type extension
		- [x] Set error type set type tokens            
		- [x] Change alias for error type
		- [x] Change alias for error type set
- [x] Implement visualizations
	- [ ] ~Implement mechanism for business object providers or business object handlers to contribute dependent objects. For now, just assume that an object with the same container is being contributed. Use relative references. This is needed to ensure that things like the diagram element for keywords exist when they are referenced.~
	- [x] Keyword Propagation Point(Internal enumeration) - Classifiers and Subcomponents.  Shape TBD. These are elements for keywords that "any", "access, "binding", etc that act similar to propagation points.
	- [x] Propagation Point - Classifiers and Subcomponents. Shape TBD.
	- [x] Propagation and Containment - Keyword Propagation Point, Feature, Propagation Point - Label
	- [x] Error Propagation Paths - Classifiers and Subcomponents - Connection. This only includes explicitly modeled error propagation paths.
	- [x] Error Flow Declaration - Classifiers and Subcomponents - Connection
	- [x] Investigate layout issues. Layout of cycles for error flow declaration. All -> All. 
- [x] Creation and editing
	- [x] Propagation Point - Renaming, Deletion, Palette.
	- [x] Propagation and Containment - Palette. Deletion. Property Section on propagation point. Consider. Adjusting flags will update references since they do not have names.
	- [x] Error Propagation Paths - Renaming, Deletion, Palette
	- [x] Error Flow Declaration - Renaming, Deletion, Palette and Property Section
	- [x] Automated tests
- [x] Other
	- [x] Consider adding tooltips
	- [ ] ~Ensure copy and paste works for EMV2 elements.~
	- [x] Consider adding diagram types. Error Type Library, Error Flow, etc.
	- [x] Add additional filters as needed. Could track whether a keyword and are referenced and have a referenced keyword filter?
	- [x] Review labels for filters
	- [x] Review palette
	- [x] Review state machine support. While state machine support is outside the scope of this task, it needs to be stable enough to be included with the plugin. Otherwise, it needs to be removed.
	- [x] Review all used graphics and annotations and refine. It would be preferred to not use "Annotation" labels as these cannot be removed and may clutter up the diagram. However, in cases of ambiguity, there needs to be a way to distinguish from exciting model elements.

	
### References
- Error Propagation and Error Containment Declarations(E.7.1)
- Error Propagation Paths and User-defined Propagation Points and Paths(E.7.3)
- Error Flow Declarations(E.7.2)

### Examples
- https://github.com/osate/examples/blob/63382782288fcc085432067d85f4ef894c858a96/EMV2TR_Test_Files/AppSystem.aadl

philip-alldredge:
Improves experimental EMV graphical support. Also provides API cleanups and additions to support graphical editor extensions.

Closes #1546


RyanMcilnay:
@philip-alldredge Creating two error types and creating an extension between the two in both directions locks the editor after trying to add another element.

1. Create error_type1
2. Create error_type2
3. Create a transition error_type1 -> error_type2
4. Create a transition error_type2 -> error_type1
5. Try to add new element to diagram

philip-alldredge:
@RyanMcilnay  does that occur if you do that in the text editor?

philip-alldredge:
@RyanMcilnay also, did you mean for it to say "transition" error type?

RyanMcilnay:
Yeah, I was about to correct myself, I meant extension.  I'll check on the initial question.

RyanMcilnay (commented):
Exception occurs when a new Error Type is added to a package diagram element and there is an existing error in the text editor such as an Error Type that extends another type that does not exist.  Just typing a new type into the text editor does not seem to cause the exception to occur.

Reproduce by:
1. Copy model below into text editor
2. Open the package diagram
3. Add a Error Type from the palette


``
package Error_Types
public
	annex EMV2 {**
		error types
			a: type extends b;
		end types;
	**};
end Error_Types;
``

[Exception.txt](https://github.com/osate/osate2/files/5277186/Exception.txt)

philip-alldredge:
@RyanMcilnay that is a serialization issue and not something we can fix. I recommend ignoring for now.

RyanMcilnay (changes requested):
The "Initial" checkbox for behavior states looks out of place, it is shifted to the right.  I think in other cases like this, we use a label called "Initial:" and a checkbox to the right with no label.

RyanMcilnay:
The "Initial" checkbox for behavior states looks out of place, it is shifted to the right.  I think in other cases like this, we use a label called "Initial:" and a checkbox to the right with no label.

RyanMcilnay:
@philip-alldredge I did not mean to close this... I thought I removing my comment

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorBehaviorTransitionHandler.java
```diff
@@ -127,16 +128,21 @@ public boolean canDelete(final CanDeleteContext ctx) {
...
 	public String getName(final GetNameContext ctx) {
-		return ctx.getBusinessObject(ErrorBehaviorTransition.class).map(bo -> {
+		return ctx.getBusinessObject(ErrorBehaviorTransition.class).map(bo -> bo.getName()).orElse("");
+	}
+
+	@Override
+	public String getNameForDiagram(final GetNameForDiagramContext ctx) {
+		return ctx.getBusinessObjectContext().getBusinessObject(ErrorBehaviorTransition.class).map(bo -> {
 			final String name = bo.getName() == null ? "" : bo.getName() + ": ";
 			return name + getText(bo.getCondition()).map(t -> "-[" + t + "]").orElse("");
```
RyanMcilnay, line 147: Name could be "-[ + t + "]->".  Since the beginning "-" is there, I think the ending arrow makes sense to have, unless there is a reason it's not there.

Also a thought, should we support changing the Event for a behavior transition from the Properties View?

philip-alldredge: Good point regarding the inconsistency. After additional consideration, I've decided to remove the condition entirely. The reason for this is that the user is likely to want to display the condition but not the name. At this time we will not support showing the condition at all. It will be available via the tooltip. 

We will not support adjusting the event at this time via the properties view. It is out of scope for this issue. The intention with regards to the state machine editing is to get the support to the point where we can release it and it won't crash and won't cause us problems when expanding support later.

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorPropagationHandler.java
```diff
@@ -0,0 +1,164 @@
...
+		return ctx.getBusinessObject(ErrorPropagation.class)
+				.map(bo -> (bo.isNot() ? "not " : "") + (bo.getDirection() == null ? "" : bo.getDirection()))
+				.orElse("");
+	}
+
+	@Override
+	public String getNameForDiagram(final GetNameForDiagramContext ctx) {
+		return ctx.getBusinessObjectContext().getBusinessObject(ErrorPropagation.class)
+				.map(bo -> (bo.isNot() ? " not " : "") + (bo.getDirection() == null ? "" : (bo.getDirection() + " "))
+						+ "propagation " + getLabelForTypeSet(bo.getTypeSet()))
```
RyanMcilnay, line 151: I don't know if I like the label showing the type set.  It can get very long and doesn't seem like there is a way to hide the label.  In similar cases like with the Error Type Set element, the user can just click on the element and view the type set in the Properties View.  I think I like that approach where the user can click on the "out propagation" or "in propagation" label, then view the type set in the Properties View.  The Properties View already shows the type set for the propagations.

Just my opinion, I can be swayed on this stance, just something to think about.

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/ErrorPropagationHandler.java
```diff
@@ -0,0 +1,164 @@
...
+	public String getName(final GetNameContext ctx) {
+		return ctx.getBusinessObject(ErrorPropagation.class)
+				.map(bo -> (bo.isNot() ? "not " : "") + (bo.getDirection() == null ? "" : bo.getDirection()))
+				.orElse("");
+	}
+
+	@Override
+	public String getNameForDiagram(final GetNameForDiagramContext ctx) {
+		return ctx.getBusinessObjectContext().getBusinessObject(ErrorPropagation.class)
+				.map(bo -> (bo.isNot() ? " not " : "") + (bo.getDirection() == null ? "" : (bo.getDirection() + " "))
```
RyanMcilnay, line 150: Remove space.  Should be "not " instead of  " not ".

RyanMcilnay (changes requested):

RyanMcilnay:
Copying and pasting Error Behavior Transitions causes exceptions.

ge/org.osate.ge.errormodel/src/org/osate/ge/errormodel/filters/PropagationPointFilter.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	@Override
+	public boolean isApplicable(final Object bo) {
+		return (bo instanceof Classifier || bo instanceof Subcomponent)
+				&& ErrorModelFilterUtil.hasApplicableErrorModelSubclause(bo);
+	}
+
+	@Override
+	public boolean test(Object bo) {
```
RyanMcilnay, line 51: A few of these test() methods need final on the paramters.

philip-alldredge:
@RyanMcilnay most of the reported issues should be resolved. I was not able to reproduce the issue with being unable to delete a flow path. Please pull and try to reproduce. If you are still unable to delete a flow path, please provide an example.

RyanMcilnay (approved):

Issue #2320, opened 2020-05-08
Move outline tree computation to background thread


lwrage:
## Summary

Outline tree computation is currently done in a foreground thread, which can block the UI. Xtext has support for updating the outline in the background.

See https://www.eclipse.org/forums/index.php/t/1103639/

related to #1832 

## Environment

* **OSATE Version**: 2.7.1
* **Operating System**: all

lwrage:
There seems to be no official documentation. Get help on the TMF forum if needed.

AaronGreenhouse:
So actually, are talking about the AADL textual outline or the AAXL instance model outline or both?

AaronGreenhouse:
Okay comments on `DefaultOutlineTreeProvider` say this:

```
/**
 * The default implementation of an {@link IOutlineTreeProvider}.
 * If you would like to run your outline in Background, please have a look at {@link BackgroundOutlineTreeProvider}.
 * Please note that this would have implications to you LabelProvider implementation. Please read the JavaDoc in the mentioned class carefully.
 * 
 * @author Jan Koehnlein - Initial contribution and API
 */
```

and `BackgroundOutlineTreeProvider` says this:

```
 * It is essential that the {@link ILabelProvider} implements {@link ILabelProviderImageDescriptorExtension} and that
 * the method {@link ILabelProviderImageDescriptorExtension#getImageDescriptor(Object)} does not need to be executed in
 * the {@link Display} thread, e.g. does not create {@link Image} objects internally.
```

This matches up with the comments in https://www.eclipse.org/forums/index.php/t/1103639/ that stress the importance of using `ImageDescriptor`.

lwrage:
Outline for the AADL text editor.

AaronGreenhouse:
So need to change `Aadl2OutlineTreeProvider` to extend `BackgroundOutlineTreeProvider`, but first I need to fix `Aadl2LabelProvider` to also extend `ILabelProviderImageDescriptorExtension`.

AaronGreenhouse:
Ugh.  SO despite what the comments on `DefaultOutlineTreeProvider ` suggest, you cannot just swap `BackgroundOutlineTreeProvider` for `DefaultOutlineTreeProvider ` (modulo the image descriptor issue) because of this nonsense:

```

	protected PolymorphicDispatcher<Void> createChildrenDispatcher = PolymorphicDispatcher.createForSingleTarget(
			"_createChildren", 2, 2, this);

[...]

PR #2454, opened 2020-09-22
Created new outline tree provide that extends BackgroundOutlineTreeProvider

AaronGreenhouse:
Fixed Issue #2320


lwrage:
Annexes should also have background outline tree providers.

AaronGreenhouse:
Updated the error model annex outline provider.  Aadl2 provider logs an error message if an annex provider is not "background".

lwrage (approved):

PR #2470, opened 2020-10-02
Improved the performance of the AADL Navigator with large models

### LINKED ISSUE #2430: Expanding AADL element in the AADL naviator is very slow
## Summary

Expanding a public package section with 100 classifiers takes more than 10s (when done for the first time after restart). Expanding a system instance with 1000 modes takes longer than I was willing to wait.

Investigate what takes so long.

We should also limit the number of child nodes shown in the navigator and generally exclude SOMs.

It may be that we read a resource again and again during expansion. Also need to check if we hold on to resources. I notices significant memory usage when expanding a large instance model.

## Environment

* **OSATE Version**: 2.8.0

AaronGreenhouse:
Fixes #2430


lwrage (commented):
You are creating SWT Image objects, which must all be explicitly disposed. Are you sure this is happening in all cases?

core/org.osate.aadl2.modelsupport/src/org/osate/aadl2/modelsupport/EObjectURIWrapper.java
```diff
@@ -63,27 +83,139 @@
 public class EObjectURIWrapper {
```
lwrage, line 83: The old constructor was public, so it should be deprecated before removing it.

AaronGreenhouse: Put back the old constructor.

AaronGreenhouse:
> You are creating SWT Image objects, which must all be explicitly disposed. Are you sure this is happening in all cases?

This is probably messed up.  The whole Image/ImageDescriptor thing in Eclipse is hard to understand.  It's unclear when or who is supposed to dispose of images.  All I know if for sure is that an Image returned by a ILabelProvider is the responsibility of the label provider, and it seems like they are disposed when the `dispose()` method on the label provider is called, but when does that happen?  

I need to look into this more, because I'm also caching images now in the Classifier Info View, so I need to make sure that does the correct thing too.

lwrage:
The label provider created images should be disposed just fine by the view. I am more concerned about images that are turned into an image descriptor. Do they need to be disposed after the descriptor has been created?

AaronGreenhouse:
> The label provider created images should be disposed just fine by the view. I am more concerned about images that are turned into an image descriptor. Do they need to be disposed after the descriptor has been created?

No, that would be bad.  The JavaDoc from `ImageDescriptor.createFromImage()` says this " Note that disposing the original Image will cause the descriptor to become invalid."

lwrage (approved):

PR #2488, opened 2020-10-20
Added subtype display to the classifier info view

### LINKED ISSUE #2436: Classifier information view should show subtype hierarchy
## Summary

The classifier view should also show component types and implementations that extend the classifier that is currently in focus.

Currently the view only shows types and implementations that are extended by the classifier in focus, i.e., the supertype hierarchy. Subtypes should also be displayed.

Compare the Java type hierarchy view, which has type hierarchy, supertye hierarchy, and subtype hierarchy.

image

## Environment

* **OSATE Version**: 2.8.0
* **Operating System**: all

AaronGreenhouse:
Fixes #2436


lwrage (changes requested):
When link with editor is active then whenever I select a classifier or feature, etc in the view, it is cleared and redrawn. I assume it unnecessarily recalculates the content whenever the selection changes.
There seems to to leftover debug output: "Cancelled" keeps getting printed to the console.

AaronGreenhouse:
> When link with editor is active then whenever I select a classifier or feature, etc in the view, it is cleared and redrawn. I assume it unnecessarily recalculates the content whenever the selection changes.

This puzzled me at first.  This is happening because we use a marker to jump to the source text.  What I didn't consider is that adding and removing the marker causes resource change events. These events are tagged with "marker change" flags.  There are a bunch of other "minor" change types as well.  I need to figure out which ones really matter.

> There seems to to leftover debug output: "Cancelled" keeps getting printed to the console.

Oops.  Fixed that.

PR #2504, opened 2020-11-04
Cleanup GE command enablement.

### LINKED ISSUE #2499: Graphical editor uses inefficient property tester
## Summary

Plugin org.osate.ge has a property tester (class `org.osate.ge.internal.expressions.ResourcePropertyTester`) to test if a file is an AADL package by loading the file into a resource set. This is horribly inefficient.

The plugin should use `org.osate.ui.handlers.AadlFileTypePropertyTester`, instead. (We should put this one in a different package!)

## Environment

* **OSATE Version**: 2.9.0
* **Operating System**: all

### LINKED ISSUE #2500: Graphical editor unnecessarily parses AADL file to enable handlers
## Summary

When right clicking on an AADL in the AADL navigator method `setEnabled()` of classes `org.osate.ge.internal.ui.handlers.OpenDiagramHandler` and `...CreateDiagramHandler` is called. In both cases `org.osate.ge.internal.ui.util.SelectionUtil.getDiagramContext(ISelection, IEditorPart)` parses the AADL file. IIUC this is done to check if a package file is selected. This should be simplified to not parse the selected AADL file.

Similarly, right click on an aaxl2 file reads the instance file into memory. Maybe it is sufficient to just check the file extension.

## Environment

* **OSATE Version**: 2.9.0
* **Operating System**: all

philip-alldredge:
Simplify enablement and visibility checks for open and create diagram commands. Closes #2499 and closes #2500.

It removes the property tester which is not needed and relies solely on a `setEnabled` that doesn't parent model files. This clarifies the behavior of the handler. The complicated `visibleWhen` check is not useful because eclipse will execute the `setEnabled` function for handlers which fail the visibility check. This essentially runs the enablement checks twice.


philip-alldredge:
@lwrage  this is ready for your review. I believe this is the best way to resolve both issues.

lwrage (approved):

lwrage (commented):
Now "create diagram" and "open diagram" show up for property sets. This introduces a new inconsistency in the UI because we distinguish them for other commands.

lwrage:
I'll soon merge PR #2426, which has some useful definitions for use in visibleWhen.

philip-alldredge:
Okay. I can re-add the visibility check using one of those properties or the one provided by AadlFileTypePropertyTester.  It'll be next week before I can revisit this.

lwrage:
See https://github.com/osate/osate2/wiki/Standard-Expression-Definitions for the new reusable definitions. These can be used in plugin.xml in enableWhen and visibleWhen.

lwrage (changes requested):
Use new definitions from org.osate.ui

lwrage (approved):

PR #2508, opened 2020-11-06
Align annex text to core.

### LINKED ISSUE #2507: Annex text not aligned to core text when edited with the graphical editor
When an annex library or subclause is edited with the graphical editor and the text editor is open, then the annex is formatted, but not aligned properly to the core text. This means that the actual annex text is formatted, but not the whitespace between `{**` and the start of the annex text and also not between the end of the annex text and `**}`. Also, the annex text itself is only indented in relation to itself and not in relation to the containing classifier or package. This can lead to a text file that looks like this:
```
package pkg1
public
  system s
    features
      f: out data port;
    annex EMV2 {**error propagations
  f: out propagation {pkg1::textually_entered_type};
end propagations;

propagation paths
  s_new_propagation_point: propagation point;
  s_new_propagation_point4: propagation point;
  s_new_propagation_point5: propagation point;
  s_new_propagation_point6: propagation point;
  s_new_propagation_point7: propagation point;
end paths;
**};
  end s;

  annex EMV2 {**error types
  textually_entered_type: type;
  new_error_type: type;
  new_error_type2: type;
end types;**};
end pkg1;
```

This occurs because `DefaultAadlModificationService` unparses the library or subclause and then simply appends `{**` to the beginning and `**}` to the end. This class should be modified to add appropriate line breaks and indentation.

joeseibel:
Fixes #2507.

### DIFF
--- a/ge/org.osate.ge/src/org/osate/ge/internal/services/impl/DefaultAadlModificationService.java
+++ b/ge/org.osate.ge/src/org/osate/ge/internal/services/impl/DefaultAadlModificationService.java
@@ -65,6 +65,13 @@ import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.IEditorReference;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.statushandlers.StatusManager;
+import org.eclipse.xtext.formatting2.FormatterPreferenceKeys;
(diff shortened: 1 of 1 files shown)


philip-alldredge (approved):

philip-alldredge:
@lwrage was is decided to resolve this issue in another way?

PR #2520, opened 2020-11-12
2442 separate flows

asazonova:
Separate flow path, sink and source. Do not show specification and implementation text in label.


lwrage (changes requested):
The PR should just address the labels for flows, but it also changes other things. Please create a new branch that just deals with the flow labels, and then create a new PR based on that branch. 

Also, please make the change in org.osate.aadl2.edit. There's a getText method in the ...Provider classes. That way the improved labels will show up in all views that use the text labels from there, not just the outline view. Don't forget to mark modified methods there with @generated NOT.

asazonova:
created a new pr with new branch for this issue

Issue #2318, opened 2020-05-05
Bidirectional feature group connection produces incorrect connection instance


thari:
## Summary

![image](https://user-images.githubusercontent.com/2693758/81102120-c3f44d00-8ed4-11ea-90ec-f8e4dd3f0139.png)

Osate produces connection instances with both source and destination being the same port.

  
## Expected and Current Behavior

Osate instantiation of a system with bi-directional feature group connections between parent and sub components produces the backward connection instance incorrectly. It is supposed to make a connection instance from a leaf level component out port to the top level component's out port. However, Osate connects back to the leaf level components out port. 

## Steps to Reproduce

1. Instantiate the ```toplevel.i``` system from the provided model
2. Check for the connection instances 
3. The second connection instance's source and destination are same

```aadl
package FeatureGroupTest
public
	
	feature group abstract_fg
		
	end abstract_fg;
	
	feature group refined_fg extends abstract_fg
		features 
			test_feature1 : in data port;
			test_feature2 : out data port;
	end refined_fg;
	
	system abstract_toplevel
		features 
			fg_port : feature group abstract_fg;
	end abstract_toplevel;
	
	system toplevel extends abstract_toplevel
		features
			fg_port : refined to feature group refined_fg;
		properties
			Classifier_Substitution_Rule => Type_Extension; --HT
	end toplevel;
	
	system implementation toplevel.i
		subcomponents
			sub : process subsystem.i;
		connections
			conn:  feature group fg_port <-> sub.fg_port;
	end toplevel.i;
	
	process abstract_subsystem
		features 
			fg_port : feature group abstract_fg;
	end abstract_subsystem;
	
	process subsystem extends abstract_subsystem
		features 
			fg_port : refined to feature group refined_fg;
		properties
			Classifier_Substitution_Rule => Type_Extension; --HT
	end subsystem;
	
	process implementation subsystem.i
		subcomponents
			th : thread subsub;
		connections
			conn2 : feature group fg_port <->th.fg_port;
	end subsystem.i;
	
	thread subsub
		features 
			fg_port : feature group refined_fg;
	end subsub;
	
end FeatureGroupTest;
```

### Other Observation

1. Even when the connection instance of two different ports are made, the connection reference with same source and destination exists 

2. Bug exist with and without the feature group refinement  

## Environment

* **OSATE Version**: 2.7.0, 2.7.1 and the latest master branch development environment
* **Operating System**:  Mac

AaronGreenhouse:
[...]

PR #2521, opened 2020-11-13
fix for #2318

thari:
recursive call connection segment direction should be same as the caller


AaronGreenhouse:
Going to close this without merging it.  This was an early attempt at fixing #2318, but didn't address the core problem with the feature names.  The actual fix is PR #2532

PR #2528, opened 2020-11-17
Allow instance model access without EMF transaction

### LINKED ISSUE #2448: OSATE hangs with "Synchronizing EMF models"
## Summary

This happens with larger models after re-instantiation when the instance editor is open. It is caused by use of the EMF transactional model editor, which manages a resource set that is automatically updated if models change. Using this editor is a leftover from early OSATE2 before the switch to Xtext. It can be removed.

Note: check that the instance editor resource set doesn't keep stale resources loaded.

## Environment

* **OSATE Version**: 2.8.0
* **Operating System**: all

lwrage:
fixes #2448


core/org.osate.aadl2.model.editor/.gitignore
```diff
@@ -1,3 +1 @@
-/bin
-/target
```
joeseibel, line 2: Was this intentional? It doesn't make sense to remove `target` from `.gitignore`.

joeseibel (changes requested):
It seems like the wrong thing to remove the `target` directory from `.gitignore`. Please change or explain that this was intentional and why it makes sense.

core/org.osate.aadl2.model.editor/.gitignore
lwrage: That was by accident.

joeseibel (approved):

PR #2529, opened 2020-11-17
2486 do not expand outline nodes without children

### LINKED ISSUE #2486: Components without children are marked as having children in the outline
## Summary

Need to check. This seems to happen for components that extend another components.

## Environment

* **OSATE Version**: 2.8.0
* **Operating System**: all

asazonova:
fixes #2486 

used imported examples to test


core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
```diff
@@ -23,6 +23,8 @@
  */
 package org.osate.xtext.aadl2.ui.outline;
 
+import javax.lang.model.type.ErrorType;
```
lwrage, line 26: This seems wrong

core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
```diff
@@ -140,7 +146,9 @@ protected boolean isLeaf(final EObject modelElement) {
 				|| modelElement instanceof FlowSpecification || modelElement instanceof FlowImplementation
 				|| modelElement instanceof EndToEndFlowImpl || modelElement instanceof Property
 				|| modelElement instanceof PropertyConstant || modelElement instanceof PropertyType
-				|| modelElement instanceof Connection) {
+				|| modelElement instanceof DataType || modelElement instanceof AbstractSubcomponent
+				|| modelElement instanceof Connection || modelElement instanceof ErrorType) {
```
lwrage, line 150: Remove ErrorType here. That should be handled in the EMV2 annex

core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
```diff
@@ -140,7 +146,9 @@ protected boolean isLeaf(final EObject modelElement) {
 				|| modelElement instanceof FlowSpecification || modelElement instanceof FlowImplementation
 				|| modelElement instanceof EndToEndFlowImpl || modelElement instanceof Property
 				|| modelElement instanceof PropertyConstant || modelElement instanceof PropertyType
-				|| modelElement instanceof Connection) {
+				|| modelElement instanceof DataType || modelElement instanceof AbstractSubcomponent
```
lwrage, line 149: Here you are saying that data types and abstract subcomponents should never be expanded. Why?

core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
```diff
@@ -168,8 +176,19 @@ protected boolean isLeaf(final EObject modelElement) {
 			return modelElement.eContainer() instanceof RecordValue;
 		} else if (modelElement instanceof IntegerLiteral) {
 			return false;
-		} else {
-			return super.isLeaf(modelElement);
+		}
+		else {
+			return !Iterables.any(modelElement.eClass().getEAllContainments(), containmentRef -> {
+				if (containmentRef.getEReferenceType() == Aadl2Package.eINSTANCE.getRealization()
```
lwrage, line 182: I would introduce a variable for `containmentRef.getEReferenceType()` to make this easier to read.

lwrage (changes requested):

core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
asazonova: i think this was supposed to not expand elements like gps1 in integration.aadl in speed regulation example. But it looks like it is unnecessary - outline view does not show them as expanded, so it is already handled by other code check

core/org.osate.xtext.aadl2.ui/src/org/osate/xtext/aadl2/ui/outline/Aadl2OutlineTreeProvider.java
asazonova: I had a hard time trying to find where to add this to in emv2 ... it seems like everything under emv2 is related to analysis. Do you have an idea where I should look? I can look again too and see if I can find it

lwrage: The org.osate.xtext.aadl.errormodel.* plugins contain the code generated from the Xtext grammar. Look in the ...ui plugin.

asazonova:
Lutz, the only other element I see that is still being shown with expand icon is this annex emv2. I think you have said before that {** **} notation allows to enter free text, so there is no way to know/control the structure of it (which makes me think that we can ignore this case in terms of this issue 2486)

device implementation aoasensor.vendorA
		annex emv2 {**
			properties
				EMV2::OccurrenceDistribution => [
					ProbabilityValue => 1.5E-5;] applies to reading;
		**};
	end aoasensor.vendorA;

lwrage (approved):

PR #2532, opened 2020-11-20
Fixed ComponentInstanceImpl.findXXXInstance methods

### LINKED ISSUE #2318: Bidirectional feature group connection produces incorrect connection instance

AaronGreenhouse:
Fixes #2318 

The `findXXXInstance()` methods didn't account for refined elements correctly.  Additionally, some of them tried to work around the problem by using name matching, but that caused other problems.


core/org.osate.aadl2/src/org/osate/aadl2/instance/impl/ComponentInstanceImpl.java
```diff
@@ -993,6 +983,113 @@ private static boolean isSameOrRefined(Connection firstconn, Connection secondco
...
+		while (rsub != null) {
+			if (e2e1 == rsub) {
+				return true;
+			}
+			rsub = rsub.getRefined();
+		}
+		return false;
+	}
+
+//	private static <T> boolean isSameOrRefined(T firstconn, T secondconn, Function<T, T> getRefined) {
```
joeseibel, line 1071: It is a little confusing that a generic version of `isSameOrRefined` is left in, but commented out. Was this supposed to be deleted, but wasn't? Were the other versions of `isSameOrRefined` supposed to call this one? Either this comment out method should be deleted or it should be used.

joeseibel (changes requested):
There is a commented out method in `ComponentInstanceImpl` which is causing confusion. Either it should be deleted or it should be used.

core/org.osate.aadl2/src/org/osate/aadl2/instance/impl/ComponentInstanceImpl.java
AaronGreenhouse: The commented-out generic version was meant to be removed.  I wanted to use it initially to avoid the duplication of the isSameOrRefined method, but it didn't work out.  The problem is the `Function<T, T>` parameter, which is meant to be satisfied by a lambda expression.  But the execution environment for `org.osate.aadl2` is 1.5 because Lutz says otherwise it doesn't build properly.  So I cannot actually pass a lambda at the call sites, and I didn't want to use ugly anonymous classes.

joeseibel (approved):

PR #2538, opened 2020-12-02
Keep contributed AADL in index when closing project

### LINKED ISSUE #2537: Packages are linking to the wrong default property set resources
## Summary
Projects/packages that are created after an existing project exists are linking to the existing project's default property sets instead of the new project's property sets.

The newly created package should be linking to the default property sets that are referenced within the owning project.

## Steps to Reproduce
1. Create a project named Test1
2. Create another project named Test2
3. Create a package in Test2 named test
4. Add"with Base_Types;" in the newly created package in Test2 public section.
The new package should look like:
````
package test
public
	with Base_Types;
end test;
````
5. Right click Test1 and select "Close Project"

An error that states "Couldn't resolve reference to ModelUnit 'Base_Types'." occurs.

## Environment

* **OSATE Version**: 2.9.0

lwrage:
fixes #2537 

@AaronGreenhouse Could you check if this works correctly with overridden contributions, please?


AaronGreenhouse (approved):
I think I may have been the one who removed this to begin with....

This fix works for me, following the steps of the original issue.  

I tried this with overridden contributed property sets.  I didn't run into any problems either with or without the fix.  This is probably because messing around with the contributed property sets causes the workspace to rebuild.

PR #2539, opened 2020-12-03
2536 import more examples

### LINKED ISSUE #2536: import additional example models
TBD 12/2/2020

1. resource budget example shows System Impl MySystem.Tier0 Â -- includes weight and power (even though  Â is not in original markdown). Markdown has  System Impl MySystem.Tier0 &nbsp; -- includes weight and power

2. same as p.1 for Tier 1
3. README for basic examples or overall for all examples
4. organize examples in proper categories
5. There are various open questions: How to best organize examples if an example can fit in several categories? What are appropriate categories? (having one project as part of several examples, but keeping only a single copy in the plugin.)
6. fix bug in the speed regulation example

meeting notes 12/2/2020
- examples can belong to multiple categories (categories should be separated and referenced by id in plugin.xml)
- all examples that are checked in [wiki](https://wiki-int.sei.cmu.edu/confluence/display/AADL/OSATE+Examples ) should be part of 2.9.1. release 

to be discussed in next meeting 12/16/2020
- review imported examples and assign proper category
- talk about replacing existing tests with JUnit regression test ( to test AADL models)
- walk through latency case study example

asazonova:
fixes #2536 

Update readme in existing examples (replace pdf links with a page link to sei resource library)

Add examples:
- Continuous Safety Risk Evaluation
- Safety Tutorial
- Stepper Motor

Example categories has been assigned per [wiki page](https://wiki-int.sei.cmu.edu/confluence/display/AADL/OSATE+Examples)


asazonova:
Lutz, I removed html folders and moved images back under markdown folder as you've asked. I also added doc/html in ignore files as a separate commit

asazonova:
Also removed svg images since they were not referenced in README file

PR #2547, opened 2020-12-14
1383 use analysis result in power requirements analysis

### LINKED ISSUE #1383: Update resource budget analyses to use AnalysisResult -> power requirements
1. Clear the mode states for a given SystemInstance. That is, the modal adapters for the model all have their modes set to null.

- Just walk the whole tree and clear the modes of all the components. We used to use the SOMs to do this so that we only touched those components that were modal. But we cannot be sure that any single SOM refers to all the modal components because some modal components may not exist in some SOMs.

2. Lookup a non-modal numeric SEI PowerCapacity property value scaled to the given unit (from platform:/plugin/org.osate.contribution.sei/resources/properties/SEI.aadl). Units used: KW, mW and W

3. Collect budget and supply information

- Supply in form of power budget drawn this power supply from other supply. This could be a requires bus access, or an incoming abstract feature. There must be a connection on this feature

If power supply > 0
- List connection info 
Else // warning unconnected power requirement (in current code, there is no code in else statement, just this comment)

Get destination connection instances for this feature
	// incoming connections: does the other end provide power?
If supply > 0, add text to report (for example) : 40.0 W from EPSU

Same for outgoing connections - if budget > 0, add text to report
	// outgoing connection. Does the other end have a power budget?
		
After you checked all features, 
	// power supply and budget based on access connections to this bus
	// we are checking whether there are connections with the component with power capacity as source or destination
	// this could be a bus, possibly an abstract component

— same logic for in/out connections again (		// Outgoing / Incoming from Power system as bus) 
       Record budget and supply

4. Building report
    Check if budget exceeds capacity
    Check if budget exceeds supply (available)

5. Write out to Excel file

asazonova:
Rebuild power requirements analysis to use analysisResult object. Method that produces an excel file also uses analysisResult object

fixes #1383


asazonova:
@AaronGreenhouse I'm confused. Jenkins seems to point to failing tests in org.osate.analysis.resource.budgets.tests. But when I run all those tests locally, they pass. My local branch is up to date ... Is there something else I am missing?

AaronGreenhouse:
As discussed over Slack/e-mail this issue needs further work to further revise the internal model, and to further restructure the class hierarchy of analysis classes and command handler classes.  These will be addressed in future meetings.

An immediate issue is that the help text shows up in the table of contents, but the link doesn't work.  This is because the HTML for the help text is missing.  It is not automatically generated from the markdown, but needs to be manually added to the project in the `help/html` directory.  This is a "shadow" of the `help/markdown` directory, so any images you add `markdown` need to be copied here as well.  

I've been using MacDown <https://macdown.uranusjr.com/> to generate HTML from the markdown file.

AaronGreenhouse:
Actually this is wrong.  The project is supposed to autogenerating the html file.  I'm not sure why this isn't working.

> An immediate issue is that the help text shows up in the table of contents, but the link doesn't work. This is because the HTML for the help text is missing. It is not automatically generated from the markdown, but needs to be manually added to the project in the `help/html` directory. This is a "shadow" of the `help/markdown` directory, so any images you add `markdown` need to be copied here as well.

AaronGreenhouse:
Feeling dumb: Forgot to run "Generate Help" from the eclipse run configurations.

Help text is okay.

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
```diff
@@ -1,28 +1,31 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: Resource Analysis Plug-in
+Bundle-Name: %pluginName
```
lwrage, line 3: Revert this change

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
```diff
@@ -1,28 +1,31 @@
...
 Bundle-SymbolicName: org.osate.analysis.resource.budgets;singleton:=true
-Bundle-Version: 4.0.3.qualifier
+Bundle-Version: 4.1.0.qualifier
 Bundle-ClassPath: .
 Bundle-Activator: org.osate.analysis.resource.budgets.ResourceBudgetPlugin
-Bundle-Vendor: CMU/SEI
+Bundle-Vendor: %providerName
```
lwrage, line 8: Revert this change

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
```diff
@@ -1,28 +1,31 @@
...
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.ui;bundle-version="[3.115.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.17.0,4.0.0)",
- org.eclipse.emf.ecore;bundle-version="[2.20.0,3.0.0)",
+ org.eclipse.emf.ecore;bundle-version="[2.20.0,3.0.0)";visibility:=reexport,
```
lwrage, line 12: Avoid visibility:=reexport unless there's a good justification for it.

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
```diff
@@ -1,28 +1,31 @@
...
  org.eclipse.core.resources;bundle-version="[3.13.0,4.0.0)",
- org.osate.aadl2;bundle-version="[4.0.0,5.0.0)",
+ org.osate.aadl2;bundle-version="[4.0.0,5.0.0)";visibility:=reexport,
```
lwrage, line 14: Avoid visibility:=reexport unless there's a good justification for it.

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
```diff
@@ -1,28 +1,31 @@
...
  org.osate.ui;bundle-version="[6.1.0,7.0.0)",
  org.osate.aadl2.modelsupport;bundle-version="[6.0.0,7.0.0)",
  org.osate.xtext.aadl2.properties;bundle-version="[3.0.0,4.0.0)",
  org.osate.analysis.architecture;bundle-version="[2.0.0,3.0.0)",
  org.osate.analysis.flows;bundle-version="[5.0.0,6.0.0)",
  org.osate.results;bundle-version="[2.0.0,3.0.0)",
- org.osate.analysis.model;bundle-version="[1.0.0,2.0.0)"
+ org.osate.analysis.model;bundle-version="[1.0.0,2.0.0)";visibility:=reexport
```
lwrage, line 21: Avoid visibility:=reexport unless there's a good justification for it.

core/org.osate.help/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.osate.help;singleton:=true
-Bundle-Version: 2.10.0.qualifier
+Bundle-Version: 2.10.1.qualifier
```
lwrage, line 5: This version must be 2.10.0 (even if a version error is displayed). It should match the release version.

lwrage (changes requested):
I found a couple of little things that could be improved.

core/org.osate.help/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

analyses/org.osate.analysis.resource.budgets/META-INF/MANIFEST.MF
asazonova: done in commit 201a300

PR #2561, opened 2021-01-22
2288 configuration flag for deactivating an annex

### LINKED ISSUE #2288: Configuration flag for deactivating an annex
In some cases, one may want to deactivate an annex from the regular parsing phase. For instance apply analyses that do not depend on them

One may imagine a configuration parameter for that, that would simply deactivate the corresponding registered parser

An open question could be: how about analysis? We could have a warning if EMV2 is disabled when trying to run FHA or FTA

asazonova:
fixes #2288 

User should be able to turn off an annex parser per specific project or workspace as a whole if needed. Each project is treated as a **separate** unit in terms of setting an annex parser preference for it. Project dependencies do not change how the preferences are set. 

Add Workspace and Project specific preference page
Add method to AnnexRegistry to list all installed annexes (needed to be shown on the preference page)
Edit AnnexParserAgent to check annex preferences before invoking the parser

Tested with projects from import example wizard


lwrage (changes requested):
Just a couple of quick initial comments

asazonova:
todo: forgot copyright info in AnnexModel.java

core/org.osate.ui/src/org/osate/ui/internal/annex/AnnexPlugin.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.ui.plugin.AbstractUIPlugin;
+import org.osgi.framework.BundleContext;
+
+/**
+ * The main plugin class to be used in the desktop.
+ * @since 6.0
+ */
+public class AnnexPlugin extends AbstractUIPlugin {
```
lwrage, line 38: What's the purpose of this class?

asazonova: I do not believe this is actually used. It was modeled after other page that had a same class. I do not believe I've seen any calls to it though in the other page so it is not needed. I'll add a todo note for myself to delete this class

asazonova:
@lwrage Lutz, eclipse is also showing the following error: (at some point there have been changes in that manifest, but right now it matches the master branch (i just compared them manually and they are the same)). I wonder if I need to revert the file for eclipse to recognize it as unchanged or should I ignore this error?

Version must be increased to 3.1.1 because the project's contents have changed	/org.osate.annexsupport/META-INF	MANIFEST.MF	line 5	Version Problem

lwrage:
The error message means that the content of the plugin jar file is different from the plugin jar file in 2.9.1. For example, you added classes, so there will be additional class files in the plugin jar.

asazonova:
@lwrage I stepped through Issue2295Test in package org.osate.core.tests.issues;    and it looks like it is not able to retrieve workspace preferences when it comes to tests. When test is executed, preference store .contains method returns false (even though I know that the preference I am testing is set and is in the preference store when UI is ran). Jenkins shows the tests as success because code ends up using a default value (annex is used, so method returns true)

For UI purposes, I have compared the way that preferences are saved and retrieved with other preference pages and those methods are identical. I have tested UI as well and it is behaving as expected (preferences work)

lwrage:
Tests run in a separate workspace. You'd have to change the preference as
part of the test to get a non default value.

asazonova:
Are you suggesting that tests should be changed to use the same workspace as UI in general?
 **Or**
 are you validating the fact that default value is expected since workspace is different?

Looking at test 2295, i would assume i can change -> testHelper.parseFile(PATH + "Issue2295.aadl") <- that AadlPackage object to UI workspace. If that's not possible, i might have to alter either ValidationTestHelper or TestHelper and the methods used there. Just wanted to clarify

lwrage:
I was just offering an explanation.

asazonova:
@lwrage, I've added help text to the annex preference page (I only committed md file so far. When I ran 'generate help', it did not generate html file for it, so I assume I will need to commit an html file as well, is that correct?)

Also, preference page per project does not show the annex help topic/article. I didn't see any of our code use this functionality. I've read that ui.dialog.propertypage should be able to show the specific text article with performHelp method (just like the workspace preference page does), but the method does not get called in debug at all and I am not finding any solutions on internet about it. So right now annex help page only shows up in workspace preference page

Also, I checked the images and they were already at 72 pixels size. Let me know if that's still too big and you want them smaller

PR #2562, opened 2021-01-24
Remove multiple consecutive keyword rules

### LINKED ISSUE #2557: Remove multiple consecutive keyword grammar rules
A while ago, we added support for displaying multiple consecutive keywords in content assist. However, this required extracting these keywords into their own grammar rules which made the grammar more ugly and difficult to read and it caused problems for formatting. We already removed these keyword rules for EMV2 in a9a9c3fede9196f2196cc9a84caa8cfb19e7bb15. The same should be done for the `Aadl2` and `Properties` grammars.

joeseibel:
Fixes #2557.


lwrage (commented):
Unfortunately this change breaks Resolute so we'll have to coordinate with Collins to merge this.

lwrage:
Superseded by #3142, which recreates this change on current `master`.

This branch was 1926 commits behind and could not be merged; most of its diff was regenerated ANTLR output and 2021-era version bumps, and several files it touched no longer exist. Rather than resolve that, the change was redone from scratch and cross-checked against this branch: the 35 keyword rules identified in `Aadl2.xtext` are exactly the 35 removed here, and the resulting grammar differs from this one by only 15 lines, all unrelated evolution since 2021.

The Resolute concern raised in the review above is unresolved and is carried forward in #3142.

PR #2570, opened 2021-02-10
2567 tests fast fail

### LINKED ISSUE #2567: New JUnit tests sometimes fail on Jenkins
**Please describe the problem your proposed feature would address**
New tests developed on Windows or Mac sometimes work fine during development but fail on Jenkins.

**What is the impact of the problem?**
Extra effort to remember the cause (case insensitive file system during development, case sensitive file system on build server) and fix it.

**Describe the solution you'd like**
Tests should fail during development if model files are loaded using a path where capitalization doesn't exactly match the file system path even on Windows and MacOS.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

lwrage:
fixes #2567

### DIFF
--- a/core/org.osate.testsupport/src/org/osate/testsupport/TestHelper.java
+++ b/core/org.osate.testsupport/src/org/osate/testsupport/TestHelper.java
@@ -151,7 +151,9 @@ public class TestHelper<T extends EObject> {
 			// test if capitalization in filePath matches capitalization in file system
 			// necessary to assure tests don't fail on Linux while working on MacOS and Windows
 			String path = new File(filePath).toPath().toString();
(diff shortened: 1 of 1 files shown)


asazonova:
Tested on mac. Tried following tests under org.osate.core.tests.issues and tests ran successfully:
- Issue1671Test.java
- Issue1831Test.java
- Issue2259Test.java

For example, I changed the capitalization like this 
		private static final String FILE = "org.osate.core.tests/Models/IsSuE2259/IssUe2259.aadl";

See test result below:

5F3B6D49-4409-4DDE-B900-562FA5E6477E_4_5005_c

lwrage:
@asazonova Please re-test on mac.

asazonova (approved):
retested same 3 tests. works as expected:
- test fails if path is incorrect due to capitalization)
- test success with unmodified path

PR #2600, opened 2021-03-01
2130b ge behavior annex support

philip-alldredge:
Will be removed if fails. Checking if creating a new pull request will fix issues with the other pull request for the behavior annex support work.


philip-alldredge:
@RyanMcilnay 

I made a few changes to the branch. Please review and update tests if necessary to fix any breakages.
- Added behavior specification filter
- Rename diagram type class
- Rename Behavior Annex Diagram menu item to be consistent with other naming -> Specification Diagram
- Renamed property section classes and ids
- Fixes property section for setting states as initial/final/complete.

As far as I know, the only additional issues that need to be resolved before merging is improving the testss. Below is a few items I noticed but I didn't take an exhaustive look:
- Add assertions. In many cases the test isn't verifying that an operation has completed. The tests don't need to assert everything but there should be enough that if something doesn't behave as expected, that the tests will fail. For example: the test didn't fail even though it was not able to set states as initial or complete.
- I would prefer to simplify openDiagramFromReference by avoid checking whether the diagram is there or not. It appears to be only used in one place. By having the check and the branch, the test will move forward regardless of whether the diagram exists. I feel like this is a bad thing because we should know whether we expect the diagram to exist or not and call and appropriate function/pass in the appropriate flag. If the diagram should not exists in all cases, rename it to openNewBehaviorSpecificationDiagramFromReference() and remove checks. 
- openDiagramFromReference accepts an index and does not use it. 
- openNewDiagramFromReference accepts an index and does not use it. If that is intended to be used to select the diagram type, do it by selecting via text label.
- Avoid using clickCheckboxInPropertiesView. Add a clickCheckboxWithId() that is an alias for clickButtonWithId in UiTestUtil and a clickCheckboxByIdInPropertiesView as necessary. Whenever possible avoid using indices to find something. It isn't always practical but I think in this case it is. Eventually, I'd like to replace other usages of the functions that use the indices.

RyanMcilnay:
@philip-alldredge Ready for review again.

philip-alldredge:
@RyanMcilnay please approve since I put in this pull request even though it is your branch.

RyanMcilnay (approved):

PR #2577, opened 2021-02-17
2478 suppress errors related to missing property sets

asazonova:
fixes #2478 
fixes #2288

Update images in Advanced OSATE Features help file. Relink referenced topics:
- annexes
- property sets
- contributed resources

Add sort for list of annexes in Annex Preference Page
Add checkbox to 'show warnings for unrecognized properties due to missing property sets'
Add preference page where user can configure a list of property sets to ignore errors for

PR #2610, opened 2021-03-05
Update Annex Preference Page

### LINKED ISSUE #2288: Configuration flag for deactivating an annex

asazonova:
- add sort to annex keys
- update screenshots for annex preference page
- move md to folder that is auto generating html

fixes #2288


lwrage (changes requested):
- Change help section title to "Managing Annexes"
- The built product doesn't contain the images for advanced features (they are in the wrong directory)

asazonova:
@lwrage Since I was moving images and relinking them in help file, I ended up fixing all 3 at the same time:
- contributed resources
- property sets
- annexes

since all 3 topics were referenced in the same advanced help file, I linked all of them to make sure they properly show when you click on ? to pop the help dialog. Code changes are submitted under [PR 2577](https://github.com/osate/osate2/pull/2577) so I don't think this branch is needed anymore

PR #2616, opened 2021-03-09
Restore error/warning markers and tooltips in instance editor

lwrage:


AaronGreenhouse (approved):
Looks good.  Editor has the decorations now.  I like it.

PR #2666, opened 2021-05-28
GE deactivate tools when editor is deactivated

RyanMcilnay:
Fix for tools not deactivating when editor changes.


philip-alldredge:
Will not merge. It changes behavior so that selecting an item from the outline will deactivate tools. We can discuss and I'll make appropriate changes directly in GEF branch.

PR #2670, opened 2021-06-04
2655 fix ba properties section text display

### LINKED ISSUE #2655: Behavior annex action is incorrectly displayed in graphical editor
- Open the Behavior Annex example "toy examples/multiModelPingPong"
- Open a behavior specification diagram for the behavior spec of `MultiModalPingPong::Cping_t.impl`
- Select the transition between `standby` and `busy`
- View the Action in the graphical editor and compare to the text editor.

Problems:
- Last brace is removed.
- Indention is removed.

RyanMcilnay:
Closes #2655

### DIFF
--- a/ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
+++ b/ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
@@ -23,13 +23,16 @@
  */
 package org.osate.ge.ba.util;
 
+import java.util.Iterator;
 import java.util.stream.Collectors;
 
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.ui.editor.model.IXtextDocument;
 import org.eclipse.xtext.xbase.lib.Pair;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterators;
+import com.google.common.collect.Lists;
 import com.google.common.collect.PeekingIterator;
 
 public class BehaviorAnnexXtextUtil {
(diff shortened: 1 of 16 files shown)


ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
```diff
@@ -68,6 +73,30 @@ public static int findUncommentedChar(final String str, final char delim) {
 		throw new RuntimeException("Cannot find terminating character");
 	}
 
+	private static int findClosingBracket(final PeekingIterator<Character> charPeekingIt, final String openBrackets,
```
philip-alldredge, line 76: This can just use Iterator, right? Doesn't need to be peaking iterator.

ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
```diff
@@ -48,14 +48,19 @@ public static String getText(final IXtextDocument xtextDocument, final XtextReso
 	/**
 	 * Find offset of character in string that is not commented out
 	 */
-	public static int findUncommentedChar(final String str, final char delim) {
+	public static int findUncommentedTerminationChar(final String str, final char delim) {
+		final String openBrackets = "{[";
+		final String closeBrackets = "]}";
 		final PeekingIterator<Character> charPeekingIt = Iterators
```
philip-alldredge, line 54: This is very inefficient. This involves boxing characters as `Character` and allocating a new list.  Use guava's `Lists.charactersOf(str)` to create a list which is a view of the existing string.  This will still box chars as `Character` but will avoid creating a new list and iterating through every character before your loop starts. It would be more efficient to keep things as char but this will still allow usage of `PeekingIterator`.

ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
```diff
@@ -48,14 +48,19 @@ public static String getText(final IXtextDocument xtextDocument, final XtextReso
 	/**
 	 * Find offset of character in string that is not commented out
 	 */
-	public static int findUncommentedChar(final String str, final char delim) {
+	public static int findUncommentedTerminationChar(final String str, final char delim) {
```
philip-alldredge, line 51: Add javadoc for this function which contains a description of the function, parameters, resulting return value/exception.

philip-alldredge (changes requested):

ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexXtextUtil.java
philip-alldredge: @RyanMcilnay I believe we had a miscommunication regarding my comment. The problem wasn't with `openBrackets` and `closeBrackets`. Those were fine as they were. The problem was with the collect operation. For example, that line can be replaced with something like `final PeekingIterator<Character> charPeekingIt = Iterators.peekingIterator( Lists.charactersOf(str).iterator());`

philip-alldredge (approved):

Issue #1183, opened 2018-04-25
Implement GEF based graphical editor


philip-alldredge:
The current graphical editor is based on Graphiti. Graphiti itself is based on GEF-legacy. The graphical editor contains work arounds for issues that we have encountered while implementing graphical editor features. The graphical editor will be migrated to use JavaFX and GEF for the following reasons.
- Font and DPI handling. The graphical editor has had problems with text clipping. Additionally, different system DPI result in diagrams being displayed differently. These issues have been minimized due to workarounds, but there are limitations on what can be done. The problems are exacerbated when trying to allow user configurable font sizes. The capabilities needed to resolve the issue are not available with GEF-legacy. JavaFX is a modern UI library which avoids many of these issues.
- Lack of Flexibility. There are several areas where Graphiti does not provide the flexibility to implement desired features. Examples of such features are adding additional tools to the palette, providing feedback while the user is using palette items, and highlighting that is orthogonal to diagram element color. Other functionality is documented in this issue tracker. After the migration occurs, more functionality will be included in the graphical editor and that functionality will be customizable as needed. 

Because there is an overlap in the capabilities of GEF MVC and the current graphical editor, the current plan is to build the editor out of the lower level GEF libraries which GEF MVC is based on. While avoiding the MVC layer will require implementing some functionality already provided by GEF MVC, it will allow a more straightforward and likely performant implementation. Once the consequences of not using GEF MVC have been explored by implementing basic functionality, a final decision will be made about which implementation path to take.

- [x] Basic GEF Non-MVC Proof of Concept. 
	- [x] Connections
	- [x] Single select based on clicking. Basic highlighting.
	- [x] Moving. Just basic to ensure there won't be unexpected issues.
	- [x] Resize. Just basic to ensure there won't be unexpected issues.
	- [x] Remove GEF MVC related code and dependencies.
- [x] Visualization. Consider SVG export while implementing.
	- [x] Shapes. Docked and undocked. 
	   - [x] Consider allowing bus graphic to adjust itself based on the largest dimension.
	- [x] Basic Connections
	- [x] Flow Indicators
	- [x] Connection decorators
- [x] Basic Editor 
	- [x] Display of diagram with styling
	- [x] Update diagram model based on JavaFX layout. As part of initial load? Connection labels should not be synchronized to diagram element unless manually set.
	- [x] Update diagram based on model changes. Handle incremental layout and checking and relinking diagram context.
	- [x] Update JavaFX nodes based on diagram changes.
[...]

PR #2673, opened 2021-06-15
update "Show as Image" toggle button on image change

RyanMcilnay:
The "Show as Image" toggle button is not updated when an image is set for elements.  This fixes that issue.
Part of #1183


philip-alldredge:
Closing without merging. A fix for the underlying cause will be committed to the GEF branch.

PR #2676, opened 2021-06-25
Fix revealing of diagram nodes for tests and link with editor for outline

RyanMcilnay:
@philip-alldredge I wanted to get your review since fixing the Outline link with editor required some changes we did not discuss.

### DIFF
--- a/ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
+++ b/ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
@@ -401,7 +401,7 @@ public class AgeContentOutlinePage extends ContentOutlinePage {
 		tree.setMenu(menu);
 		getSite().registerContextMenu("org.osate.ge.editor.AgeDiagramEditor", menuMgr, viewer); // Allow contributions
 
-		editor.addSelectionChangedListener(event -> updateOutlineSelectionIfLinked(false));
+		editor.addSelectionChangedListener(event -> updateOutlineSelectionIfLinked());
 		editor.getDiagram().addModificationListener(diagramModificationListener);
 
 		viewer.addSelectionChangedListener(this);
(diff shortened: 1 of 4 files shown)


ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
```diff
@@ -447,6 +451,9 @@ private void updateOutlineSelectionIfLinked() {
 				if (treeViewer != null && treeViewer.getContentProvider() != null
 						&& !outlineNodes.equals(editor.getSelectedDiagramElementSet())) {
 					treeViewer.setSelection(buildDiagramNodeTreeSelectionFromEditor());
+					if (revealNode) {
+						editor.reveal(outlineNodes.iterator().next());
```
philip-alldredge, line 455: I don't think `revealNode` and the call to `reveal` is desired. This is called to update the outline to match the editor. I don't think we want to reveal the current selection when the link with editor button is selected.

ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
```diff
@@ -484,6 +491,7 @@ private void selectEditorDiagramElements() {
 		if (getTreeViewer() != null && getTreeViewer().getContentProvider() != null
 				&& !outlineElements.equals(editorElements)) {
 			editor.selectDiagramNodes(outlineElements);
+			editor.reveal(outlineElements.iterator().next());
```
philip-alldredge, line 494: This assumes that outline nodes has at least one item selected. Will throw an exception if there is ever a case where the selection is empty. Consider outlineNodes.stream().findAny().ifPresent(...);

philip-alldredge (changes requested):

ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
RyanMcilnay: Yep, you are correct... not sure why I did that.  fixed

ge/org.osate.ge/src/org/osate/ge/internal/ui/editor/AgeContentOutlinePage.java
RyanMcilnay: Yeah, probably a  good idea.  I used findFirst since findAny has no guarantee of being the first element

PR #2686, opened 2021-07-15
2645 Display dispatch conditions on diagram

### LINKED ISSUE #2645: Allow displaying behavior annex dispatch condition on diagram
Add the ability to show the dispatch condition for a behavior transition on the diagram. This should be implemented similarly to showing the subcomponent classifier. Add filter as well. Dispatch conditions should be shown by default for behavior specification diagrams.

RyanMcilnay:
Closes #2645


ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
```diff
@@ -0,0 +1,48 @@
...
+ * aries to this license with respect to the terms applicable to their Third Party Software. Third Party Software li-
+ * censes only apply to the Third Party Software and not any other portion of this program or this program as a whole.
+ */
+package org.osate.ge.ba.businessobjecthandlers;
+
+import org.osate.ba.aadlba.BehaviorTransition;
+
+public class DispatchCondition {
+	private static final String NAME = "Dispatch Condition";
+	private static final String KEY_BEHAVIOR_TRANSITION_DISPATCH_CONDITION = "bt_dc"; // TODO add filter
```
philip-alldredge, line 30: Remove TODO

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchConditionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+	public final static String DISPATCH_CONDITION = BehaviorAnnexReferenceUtil.TRANSITION_TYPE + ".dispatch_condition";
+	private final Graphic labelGraphic = LabelBuilder.create().build();
+
+	@Override
+	public boolean isApplicable(final IsApplicableContext ctx) {
+		return ctx.getBusinessObject(DispatchCondition.class).isPresent();
+	}
+
+	@Override
+	public CanonicalBusinessObjectReference getCanonicalReference(final ReferenceContext ctx) {
```
philip-alldredge, line 59: Create a proper canonical reference. According to the documentation, this function should never return null. It is allowed to avoid problems with old code currently. See `ModeTransitionTriggerHandler` as an example of using the container/owner as part of a canonical reference.

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchConditionHandler.java
```diff
@@ -0,0 +1,104 @@
...
+		return null;
+	}
+
+	@Override
+	public RelativeBusinessObjectReference getRelativeReference(final ReferenceContext ctx) {
+		return new RelativeBusinessObjectReference(DISPATCH_CONDITION,
+				ctx.getBusinessObject(DispatchCondition.class).get().getKey());
```
philip-alldredge, line 66: What is the purpose of the key here?

philip-alldredge, line 66: Use `orElseThrow()` instead of `get()`. It does the same thing and the no-argument version is available in java 11. It's clearer that it is your intention and avoids warnings with spot bugs.

ge/org.osate.ge.ba/src/org/osate/ge/ba/diagram/contentFilters/DispatchConditionFilter.java
```diff
@@ -0,0 +1,52 @@
...
+ * censes only apply to the Third Party Software and not any other portion of this program or this program as a whole.
+ */
+package org.osate.ge.ba.diagram.contentFilters;
+
+import org.osate.ba.aadlba.BehaviorTransition;
+import org.osate.ge.ContentFilter;
+import org.osate.ge.ba.businessobjecthandlers.DispatchCondition;
+
+public class DispatchConditionFilter implements ContentFilter {
+	public static final String ID = "ba.bt.dispatchCondition";
```
philip-alldredge, line 31: This should be `ba.dispatchCondition`.

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
```diff
@@ -0,0 +1,48 @@
...
+
+	public DispatchCondition(final BehaviorTransition owner) {
+		this.owner = owner;
+	}
+
+	public BehaviorTransition getOwner() {
+		return owner;
+	}
+
+	public String getKey() {
```
philip-alldredge, line 41: What is the purpose of the key?

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
philip-alldredge, line 30: What is the purpose of the key?

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.osate.ba.aadlba.BehaviorTransition;
+
+public class DispatchCondition {
+	private static final String NAME = "Dispatch Condition";
+	private static final String KEY_BEHAVIOR_TRANSITION_DISPATCH_CONDITION = "bt_dc"; // TODO add filter
+	public final BehaviorTransition owner;
+
+	public DispatchCondition(final BehaviorTransition owner) {
+		this.owner = owner;
```
philip-alldredge, line 34: Should use Objects.requireNonNull to ensure owner isn't null.

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
```diff
@@ -0,0 +1,48 @@
...
+ * conditions contained in any such Third Party Software or separate license file distributed with such Third Party
+ * Software. The parties who own the Third Party Software ("Third Party Licensors") are intended third party benefici-
+ * aries to this license with respect to the terms applicable to their Third Party Software. Third Party Software li-
+ * censes only apply to the Third Party Software and not any other portion of this program or this program as a whole.
+ */
+package org.osate.ge.ba.businessobjecthandlers;
+
+import org.osate.ba.aadlba.BehaviorTransition;
+
+public class DispatchCondition {
```
philip-alldredge, line 28: Document why this is required. Behavior transition has a `getCondition` that returns a `BehaviorCondition`. Is that object not suitable as a business object? Other parts of the code imply that it could be null when there is a condition. There should be a comment that clearly documents the rationale for using this class. Consider implementing EmfContainerProvider (See `SubprogramCallOrder` for an example).

ge/org.osate.ge.ba/src/org/osate/ge/ba/BehaviorAnnexBusinessObjectProvider.java
```diff
@@ -49,6 +52,9 @@
 			final BehaviorAnnex ba = (BehaviorAnnex) bo;
 			return Stream.concat(Stream.concat(ba.getTransitions().stream(), ba.getStates().stream()),
 					ba.getVariables().stream());
+		} else if (bo instanceof BehaviorTransition) {
+			final BehaviorTransition bt = (BehaviorTransition) bo;
+			return getDispatchCondition(bt);
```
philip-alldredge, line 57: No need for a separate function or using streams, filtering, and mapping. Just keep it simple.  Just return `Stream.of(new DispatchCondition())` if the transition has a condition that can be shown.  The current check is checking if bt is null which is never the case.

ge/org.osate.ge.ba/src/org/osate/ge/ba/util/BehaviorAnnexUtil.java
```diff
@@ -162,4 +167,65 @@ public AadlPackage getDataClassifierPackage() {
 		final AadlPackage pkg = root instanceof AadlPackage ? (AadlPackage) root : null;
 		return Optional.ofNullable(pkg);
 	}
+
+	/**
+	 * Helper to create an {@link EmbeddedTextValue} and get text for the {@link DispatchCondition} of a specified {@link BehaviorTransition}
```
philip-alldredge, line 172: I don' think this should be inside `BehaviorAnnexUtil`. It should be it's own package level class.

Additionally, I think it would be better to make it a utility class (no instantiation) and just have the two public static methods for creating the text value or getting the text. May require adding private methods or slight refactoring to avoid some duplication. The constructor solves the problem but it makes it more awkward on the caller in my opinion

philip-alldredge (changes requested):
Additional comments:

- Don't commit the digest files
- I am getting exceptions. I think this is due to your `plugin.xml` changes not being committed.  The dispatch condition handler and filters are not being used.

ge/org.osate.ge.ba/src/org/osate/ge/ba/businessobjecthandlers/DispatchCondition.java
RyanMcilnay: I believe this class is required because we want the label in the outline to just be "Dispatch Condition".  I cannot use BehaviorCondition and have an empty name because the outline will show <Unnamed Behavior Condition>.

I will document this, but just wanted to let you know the reasoning incase I am missing something.

RyanMcilnay:
@philip-alldredge Should be ready to review again.  This also contains filtering for the ErrorModel palette commands now like we discussed in our meeting.  (Fixes commands showing up for instance diagrams for example.)

ge/org.osate.ge.ba/src/org/osate/ge/ba/diagram/diagramType/BehaviorSpecificatonDiagramType.java
```diff
@@ -58,6 +60,8 @@ public boolean isApplicableToContext(final Object contextBo) {
 	public ImmutableSet<String> getDefaultContentFilters(final Object bo) {
 		if (bo instanceof BehaviorAnnex) {
 			return ImmutableSet.of(BehaviorVariableFilter.ID, BehaviorTransitionFilter.ID, BehaviorStateFilter.ID);
+		} else if (bo instanceof BehaviorTransition) {
```
philip-alldredge, line 63: Let's avoid adding it to the default content filters for now. Connection labels layout are disabled because of an ELK exception and this will likely result in problems.

philip-alldredge (changes requested):

philip-alldredge (approved):

PR #2689, opened 2021-07-24
Ask to save open editors before getting selected component impl

### LINKED ISSUE #2613: NPE when instantiating when file not saved
**Summary**
A NullPointerException is raised when instantiating a system *but* the file does not preexist.

**Expected behavior**
A valid instance or an error

**Actual behavior**
A NullPointerException

	at org.osate.aadl2.instantiation.InstantiateModel.getInstanceModelURI(InstantiateModel.java:500)
	at org.osate.ui.handlers.InstantiationHandler$KickoffJob.run(InstantiationHandler.java:147)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

**Steps To Reproduce**
1. Create a project, and a package with a system to instantiate
2. *DO NOT SAVE THE FILE*
3. In the outline, right click on the system to instantiate it.

**Desktop (please complete the following information):**
 - OSATE Version: 2.9.0
 - Operating System:  macOS

lwrage:
The NPE happens if the component implementation that is selected has not yet been saved. In this case the EObject for the implementation is not found. The fix first asks the user to save open files and then tries to find the EObject. We ask to save, anyway, so I just moved that call up.

fixes #2613

### DIFF
--- a/core/org.osate.ui/src/org/osate/ui/internal/instantiate/AbstractInstantiationEngine.java
+++ b/core/org.osate.ui/src/org/osate/ui/internal/instantiate/AbstractInstantiationEngine.java
@@ -73,33 +73,35 @@ abstract class AbstractInstantiationEngine<T> {
 	 * XXX: Say something about the abstract methods here
 	 */
 	public final List<IFile> instantiate(final IProgressMonitor monitor) {
-		final Set<T> inputs = getInputsFromSelection(selectionAsList);
-		final int size = inputs.size();
-
 		boolean cancelled = false;
 
-		if (size > 0) {
-			/*
-			 * This map is shared by all the jobs to build the set of results. It is created here,
-			 * given to all the jobs, and then used to build te final method result.
-			 */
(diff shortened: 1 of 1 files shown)


AaronGreenhouse:
Change to the code makes sense, but sadly isn't enough to fix the problem.  Turns out that `IDE.saveAllEditors` which is used by `saveDirtyEditors` doesn't return enough information.  It returns `true` or `false` to indicate if the save was cancelled.  But files can also be skipped, and that is not indicated.  For 1 dirty file, you get a dialog with "Don't Save", "Cancel", and "Save" as options.  Selecting "Cancel" causes a `false` return value, but choosing "Save" or "Don't Save" causes a `true` return value!

Even worse, if you have multiple dirty editors, then the dialog gives you a checklist and you can choose which editors are saved or not.  Only if you select the "Cancel" button does the method return false.  It doesn't indicate at all which editors were chosen not to be saved.

PR #2719, opened 2021-10-19
add a check to support consistency rule C1 of AADL standard

### LINKED ISSUE #2652: Missing validation for reserved thread ports
**Summary**

Threads have a couple of reserved ports: The `Complete out event port`, `Error out event data port`, `Abort out event port`, and `Stop out event port` are reserved, i.e., they must be declared of the specified port type. OSATE should validate that these ports are declared correctly, but doesn't. 

See 5.4 (L3) in the AADL 2.2 standard.

**Expected behavior**
A port named `complete` should be valid only as an `out event port`, etc.

**Actual behavior**
There is no validation, e.g. `complete` can be declared as an `in data port` without an error in the model.

**Steps To Reproduce**

1. Paste model below into the AADL text editor

```aadl
package ReservedPorts
public

  thread Correct
    features
      complete: out event port;
      error: out event data port;
      abort: out event port;
      stop: out event port;
  end Correct;

  thread Invalid
    features
      -- there should be errors on each feature
      complete: in data port;
      error: in event port;
      abort: out event data port;
      stop: in event port;
  end Invalid;
  
end ReservedPorts;
```

**Desktop (please complete the following information):**
 - OSATE Version: 2.9.2
 - Operating System: Linux

asazonova:
Consistency rule C1 in AADL standard for threads:
AADL reserves the following ports. If defined, they must match the following constraints:

- Complete must be an out event port
- Error must be an out event data port
- Abort must be an out event port
- Stop must be an out event port

fixes #2652


lwrage:
Don't forget to add a test.

lwrage:
Did you test if all included example models work with this change? It is possible that one of them uses a reserved name with the wrong kind of port.

asazonova:
> Did you test if all included example models work with this change? It is possible that one of them uses a reserved name with the wrong kind of port.

image

yes, this is what i see which matches the consistency rule C1. Error messages are also consistent with the rule C1

jjhugues:
> yes, this is what i see which matches the consistency rule C1. Error messages are also consistent with the rule C1

But what about the examples from the Wizard? Could check there is no regression with them?  Thanks

asazonova:
> @asazonova I meant the example models included with OSATE (File > New > Example).

@lwrage  none of the examples in example wizard have the threads that use reserved ports out of the box. I added a reserved port as a part of connection and flow in few of the examples that had threads and it worked as expected (showed the error to specify the exact type that port should be and once corrected, the error went away). In my understanding, this consistency rule only applies to threads, so reserved ports in process were not showing the error (consistency rule was not applied to the process)

image

lwrage (approved):

PR #2730, opened 2021-11-16
2722 connection som

### LINKED ISSUE #2722: ConnectionInstances need to consider the modes of the endpoints
**Please describe the problem your proposed feature would address**

When a semantic connections exists between two components, it should only exist in the system operation modes that both components exist in.

When a semantic connection exists between a component and a feature it should only exist in the system operation modes that the component exists in.  

**What is the impact of the problem?**

Currently semantic connections (`ConnectionInstance` objects) between modal components do not have their `InSystemOperationModes` attribute set.  This attribute is only set if one of the declarative connections that makes up the semantic connection is modal.

This can lead to a view of the instance model where a semantic connection exists, but the components it is connecting do not.

Furthermore, things are complicated by the fact that `ConnectionInstance.isActive()` does take this information into account:

```
        public boolean isActive(SystemOperationMode som) {
                if (getInSystemOperationModes().isEmpty() || getInSystemOperationModes().contains(som)) {
                        if (getSource().getComponentInstance().isActive(som)
                                        && getDestination().getComponentInstance().isActive(som)) {
                                return getContainingComponentInstance().isActive(som);
                        }
                }
                return false;
        }
```

**Describe the solution you'd like**

The process that applies SOMs to connection instances need to be updated to handle this situation.

lwrage:
fixes #2722


AaronGreenhouse (approved):
Changes seem reasonable.  Works on my original examples, and on the hypothetical computer system models I  developed for safir.

PR #2740, opened 2022-01-12
New EMV2 Instance

### LINKED ISSUE #2802: Instantiate basic EMV2 elements
When instantiating a model, we should also instantiate the error model elements of that model. This should start with propagations, error flows, and propagation paths. Further elements will be handled by future issues.

joeseibel:
Fixes #2802.

This is the updated pull request and branch for the error model instance. The old pull request can be found here: https://github.com/osate/osate2/pull/2646.

This branch started with taking Peter's meta-model, instantiator, tests, and other supporting code and copying it to a new branch. Now that his code runs again, we can begin to evaluate his meta-model.

Here are some low level issues found with the meta-model:

- The type `EOperation` is poorly named. A more meaningful name should be selected and especially one that does not conflict with a standard EMF type name.
- Evaluate if the types `ConstraintElement`, `ConstraintExpression`, and `ConstrainedInstanceObject` are properly named.
- `StateMachineInstance` has the field `currentState`. This seems to be similar to how `SystemInstance` keeps track of its current SOM. Peter's instantiator sets `currentState` to the current state, but there is no code that updates this field. Do we really want to handle state by mutating the model? Would it be better for an analysis or utility method to iterate and keep track of the current state rather than storing it in the model?
- `ConstrainedInstanceObject` has the field `instanceObject` which points to an `InstanceObject`. The name of this field should be updated to be more descriptive. Also, the type should be more restrictive.
- `ConstrainedInstanceObject.propagationKind` has the type `EString`. Can this be a better type such as an enum? I think this is supposed to be set when a propagation doesn't refer to a feature, but instead is marked as `processor`, `memory`, `connection`, etc. I think that the field is a string in the declarative because of Xtext restrictions, but the instantiator should be able to convert that to an enum.
- For the type `ConstrainedInstanceObject`, it seems that either the field `instanceObject` should be set or the field `propagationKind` should be set. It seems like they will never both be set. If this is true, this restriction should be expressed in the meta-model.
- What is the relationship between `ConstrainedInstanceObject` and `ErrorPropagationInstance`. Should these be one type or does the inheritance make sense? The field `propagationKind` is on `ConstrainedInstanceObject`, but it would seem that it should be on `ErrorPropagationInstance`, but it is never set when constructing an `ErrorPropagationInstance`.
- `StateTransitionInstance` has the field `stateTransition` which points to an `EObject`. The type of this field should be more restrictive.
- `ErrorFlowInstance` has the field `emv2Element` which points to a `NamedElement`. The name of this field should be updated to be more descriptive. Also, the type should be more restrictive.
- `ErrorFlowInstance` has the fields `source` and `sink`, both of which are `EBoolean`s. First of all, these fields are never used. It seems like the intention is that `source` will be `true` for flow sources, `sink` will be `true` for flow sinks, and both will be `true` for flow paths. There are so many alternatives which could be better such as an enum, checking the fields `incoming` and `outgoing` for `null`, or even creating the subtypes `ErrorFlowSourceInstance`, `ErrorFlowSinkInstance`, and `ErrorFlowPathInstance`.
- `ErrorPropagationConditionInstance` has the field `emv2Element`. The name of this field should be updated to be more descriptive.
- `ErrorPropagationConditionInstance` has the fields `source` and `sink`, both of which are `EBoolean`s. I don't think these fields make any sense and they are probably a copy & paste error.
- `ErrorDetectionInstance` has the field `emv2Element`. The name of this field should be updated to be more descriptive.
- `ErrorDetectionInstance` has the field `errorCode` which points to an `EString`. In the declarative, this could be an integer, a string, or a property constant. I'm assuming that the property constant only makes sense when it points to an integer or a string. Currently when instantiating, Peter takes any integers and converts them to strings. Is this correct? Should we make a distinction between an integer and a string in the instance meta-model.
- `ConstraintExpression` has the field `k` which is an `ELong`. This is only used when the `operator` is `KORMORE`. Should the operator really be an enum or should we have specific subtypes with only the fields that are needed?
- `PropagationPathInstance` has the field `emv2Element` which points to a `NamedElement`. The name of this field should be updated to be more descriptive. Also, the type should be more restrictive.


joeseibel:
Fixes #2802.

sprocter:
Hi,

I'm not sure how to review this -- it "seemed" too large to me, and sure enough a quick google turned up "optimal sizes" of pull requests that are 2-3 orders of magnitude smaller than this one (see, eg, [1](https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/), [2](https://smallbusinessprogramming.com/optimal-pull-request-size/), [3](https://hugooodias.medium.com/the-anatomy-of-a-perfect-pull-request-567382bb6067)). Even granting that much of the 58,000sloc is generated, this is too large for my review.

Can it be restructured and split into multiple PRs? Or do you (@lwrage) want to waive my request?

PR #2775, opened 2022-03-28
Add acvip examples

asazonova:
add AADL examples from ACVIP youtube video


asazonova:
@lwrage i haven't done a full local build yet to see if examples show properly that way as well (only tested locally so far). i should have time to do that tomorrow.

asazonova:
@lwrage tested local build. aadl files do not show any errors and reflect new DM text (as well as html and md files have new DM text). Help info shows html file correctly

PR #2788, opened 2022-04-13
Remove xtend from textual instance

joeseibel:


joeseibel:
The formatter doesn't play nicely with text blocks in Java that contain curly braces. Need to discuss with @lwrage.

lwrage (approved):

Issue #2782, opened 2022-04-07
Bound resource allocations analysis report not generated


ericsenn:
**Summary**
Trying to generate a "bound ressource allocations analysis" report.

**Expected behavior**
A report should be generated.

**Actual behavior**
The report is not generated
It is generated if you comment the bus subcomponents declaration (see aadl code below)
in the HW platform defined as a system implementation.
There is no problem with Osate2-2.9.1

According to Lutz Wrage, from OSATE forum : "This is a bug. The analysis throws an exception and writes a stack trace in the error log. The fix should be easy. Could you file an issue in the osate issue tracker on github, please?"

**Steps To Reproduce**
Create an aadl package with the code below
Instantiate system implementation mysys.i
From the instance, run the Analysis > Budget > Analyse Ressource Allocations (bound)
The report is not generated.

Comment the ```thebus: bus mybus.i;``` in the model.
Re-instantiate
Re-run analysis
The report is generated.

```aadl
package sched_mips
public

with SEI;


--*************************************************** SW

process a_process
end a_process;

process implementation a_process.i
    subcomponents
        thread1 : thread a_thread.i;
    properties
        SEI::MIPSBudget => 1.0 MIPS;
end a_process.i;

thread a_thread
end a_thread;

thread implementation a_thread.i
properties
        Period => 10 Ms;
        Compute_execution_time => 5 Ms .. 5 Ms;
end a_thread.i;

--*************************************************** HW

processor proc
    features
        buscon: requires bus access mybus.i;
        end proc;

processor implementation proc.i
    properties
        SEI::MIPSCapacity => 2.0 MIPS;        
        end proc.i;
       
bus mybus
    end mybus;
bus implementation mybus.i
    end mybus.i;

system myhw
    end myhw;
   
system implementation myhw.i
    subcomponents
        proc: processor proc.i;
        thebus: bus mybus.i;--IF YOU COMMENT THIS LINE THE REPORT IS GENERATED
end myhw.i;

--*************************************************** SW+HW+bindings

system mysys
    end mysys;

system implementation mysys.i
    subcomponents
        sw: process a_process.i;
        hw: system myhw.i;
    properties
        Actual_Processor_Binding => (reference (hw.proc)) applies to sw;
        end mysys.i;

end sched_mips;
```

**Desktop (please complete the following information):**
 - OSATE Version: OSATE2-2.10.2
 - Operating System: Linux Ubuntu 18.04LTS

passeris:
[...]

Issue #2772, opened 2022-03-17
FTA fails on example model


lwrage:
**Summary**
Running the FTA on `wbs.ima` from the simple ARP4761 example does not produce a result. 

**Expected behavior**
There should be either a result or an error message.

**Actual behavior**
The FTA fails silently and only reports a stack trace in the error log.

**Steps To Reproduce**

1. Import the ARP4761 example model
2. Instantiate `wbs.ima` from `simple/wbs.aadl`
3. Run the FTA on the created instance model (using default settings in the dialog).
4. See error log.

**Desktop (please complete the following information):**
 - OSATE Version: 2.11 development
 - Operating System: Linux

**Additional context**
```
eclipse.buildId=unknown
java.version=17.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Framework arguments:  -product org.osate.branding.osate2
Command-line arguments:  -product org.osate.branding.osate2 -data /home/lw/local/eclipse/java-2022-03/ws/../runtime-osate2 -dev file:/home/lw/local/eclipse/java-2022-03/ws/.metadata/.plugins/org.eclipse.pde.core/OSATE2/dev.properties -os linux -ws gtk -arch x86_64 -consoleLog

org.eclipse.e4.ui.workbench
Error
Thu Mar 17 11:05:46 EDT 2022
Command 'org.osate.aadl2.errormodel.analysis.fta' failed

org.eclipse.core.commands.ExecutionException: Error executing 'org.osate.aadl2.errormodel.analysis.fta': org.osate.aadl2.properties.PropertyDoesNotApplyToHolderException: Property Actual_Function_Binding does not apply to org.osate.aadl2.instance.impl.ComponentInstanceImpl
	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.execute(HandlerServiceHandler.java:170)
	at org.eclipse.core.commands.Command.executeWithChecks(Command.java:488)
	at org.eclipse.core.commands.ParameterizedCommand.executeWithChecks(ParameterizedCommand.java:487)
	at org.eclipse.e4.core.commands.internal.HandlerServiceImpl.executeHandler(HandlerServiceImpl.java:213)
	at org.eclipse.e4.ui.workbench.renderers.swt.HandledContributionItem.executeItem(HandledContributionItem.java:438)
[...]

PR #2789, opened 2022-04-15
Handle PropertyDoesNotApplyToHolderException in generated property getters

### LINKED ISSUE #2593: Generated property getter methods don't handle the case when the property doesn't apply
**Summary**

When updating the `FlowLatencyAnalysisSwitch` to use the new generated property getter methods, I discovered that the generated property set code doesn't nicely handle the case where the property doesn't to apply to an AADL element.  The underlying property lookup code throws a `PropertyDoesNotApplyToHolderException`.  This is allowed to propagate to the caller.  This would be fine if there were an easy way to test if the property applies, but there isn't one.  You really just have to try it and see what happens.  This exception should not be propagated, but caught internally to take advantage of the `Optional` class.

**Expected behavior**

I expected the generated property getter method to return an _empty_ `Optional` instance in this case, same as when the property value is undefined.

**Actual behavior**

The underlying `PropertyDoesNotApplyToHolderException` propagates to the caller of the method.

** Possible fix **

I fixed this by updating CodeGenUtil.lookupProperty() on my local branch to be

```
	public static PropertyExpression lookupProperty(Property property, NamedElement lookupContext,
			Optional<Mode> mode) {
		try {
			Optional<PropertyExpression> modalValue = mode.map(m -> {
				PropertyAssociation association = lookupContext.getPropertyValue(property).first();
				if (association == null) {
					PropertyExpression defaultValue = property.getDefaultValue();
					if (defaultValue == null) {
						throw new PropertyNotPresentException(lookupContext, property, "No property value");
					} else {
						return defaultValue;
					}
				} else {
					return association.valueInMode(m);
				}
			});
			return modalValue.orElseGet(() -> lookupContext.getNonModalPropertyValue(property));
		} catch (final PropertyDoesNotApplyToHolderException e) {
			throw new PropertyNotPresentException(lookupContext, property, "Property does not apply");
		}
	}
```

This seems to do what I want, but I'm not sure if there are other places I need to catch the exception in the code used by the generated method.

joeseibel:
Fixes #2593.


lwrage:
@joeseibel Could you add a test for #2782, please? There is also an "applies to" check in the fault tree analysis (see #2772) that is now obsolete.

joeseibel:
The problem from #2782 was tested manually. A unit test was not created because `BoundResourceAnalysis` has not had its logic and UI code fully separated.

lwrage (approved):

PR #2951, opened 2025-08-21
2712 hot key binding conflict on macos commandi

AaronGreenhouse:
Updated key binding for Instantiate Model on MAC to be CMD+SHIFT+I.  Should still be CTRL+I on windows and linux.  

I've tested on MAC.  I don't have the ability to test on windows or linux.


AaronGreenhouse:
Added key binding for "Reinstantiate Model": M1+R.  SO should be COMMAND+R on MacOS and CTRL+R on others.  Tested on Mac.  Needs to be tested on Windows and Linux.

PR #2954, opened 2025-08-22
Updated Weight Analysis to output component's net weight

AaronGreenhouse:
Updated output info markers to add a marker for each physical component that reports the net weight.


lwrage:
Test failure: org.osate.analysis.architecture.tests.WeightTotalsTest.testWeightTotals

lwrage (approved):

PR #2956, opened 2025-08-25
Fixed so that thread and thread group components are also considered

AaronGreenhouse:
Fixed weight analysis to explicitly ignore thread and thread group components.


AaronGreenhouse:
Added test for issue 2650

lwrage (approved):

PR #2957, opened 2025-08-28
Force metaclass references to be styled as DEFAULT

AaronGreenhouse:
Metaclass references are styled as default.

Swapped out use of the deprecated class org.eclipse.xtext.ui.editor.syntaxcoloring.ISemanticHighlightingCalculator with the newer org.eclipse.xtext.ide.editor.syntaxcoloring.ISemanticHighlightingCalculator.


lwrage (approved):
looks good

PR #2959, opened 2025-09-02
Refactored CSV output classes

AaronGreenhouse:
Removed AbstractAnalysisHandler's dependency on IFile and Eclipse workspaces, and instead made NewBusLoadAnalysisHandler deal with the conversion from IFile to normal Java library abstractions.


AaronGreenhouse:
Changed the NewBusLoadAnalysis class to take care of creating the CSV file. New base class for CSV files is org.osate.result.util.AbstractCSVResultWriter.

This way the bus load analysis can be invoked from a headless OSATE instance.

lwrage (approved):

Issue #2988, opened 2026-08-04
End-to-end flow connection matching ignores declarative connection context 🤖


lwrage:
### Summary

Declarative connections in end-to-end flows are matched to connection references by unqualified name rather than model identity and context.

`CreateEndToEndFlowsSwitch.testConnection` compares `Connection.getName()` values case-insensitively. Connections in different component implementations may legally have the same unqualified name, so an unrelated nested connection can satisfy the match for a connection declared in the parent implementation.

### Reproduction

Define a top-level connection named `c` that is part of an end-to-end flow. Also define an unrelated connection named `c` inside a nested component. Add another top-level connection whose semantic connection instance continues through that nested `c`.

The nested connection name is incorrectly accepted as a match for the top-level flow connection, producing a spurious end-to-end flow instance or an instance with the wrong destination.

### Expected behavior

Connection matching should respect the referenced declarative connection identity, including refinement and containing-component context where needed. An unrelated connection with the same unqualified name must not match.

### Relevant code

`core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java`, method `testConnection`.

Issue #2987, opened 2026-08-04
Cyclic nested end-to-end flows create cyclic instance graphs 🤖


lwrage:
### Summary

Cyclic references between nested end-to-end flows are treated as successfully instantiated and can create cyclic instance graphs or unbounded recursion.

`instantiateEndToEndFlow` inserts the declarative flow into `ete2info` before processing its segments. `processEndToEndFlow` uses only `containsKey` to decide that a referenced flow is already instantiated. A reference back to a flow currently being processed therefore reuses its incomplete instance as if it were complete.

There is already a `TODO` in this path to detect cyclic dependencies between end-to-end flows.

### Reproduction

Declare two end-to-end flows where `a` contains `b` and `b` contains `a`, and reference one of them from another flow or instantiate the containing implementation.

The instantiated model can contain `EndToEndFlowInstance` objects that recursively contain each other. Code that follows the first or last nested flow element can then recurse indefinitely or fail with a stack overflow.

### Expected behavior

Instantiation should distinguish in-progress flows from completed cached flows, detect the cycle, report an instantiation diagnostic, and avoid retaining a cyclic or incomplete end-to-end flow instance graph.

### Relevant code

`core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java`, methods `instantiateEndToEndFlow` and `processEndToEndFlow`.

Issue #2986, opened 2026-08-04
Nested end-to-end flow composition combines incompatible path variants 🤖


lwrage:
### Summary

Composing a parent end-to-end flow with a nested end-to-end flow creates a Cartesian product of incompatible incoming paths and nested variants.

`ETEInfo.preConns` records the declarative connections that precede each nested flow instance. However, when the parent already has incoming semantic connection candidates, `processEndToEndFlow` combines every incoming connection instance with every nested `ETEInfo` without checking `preConns`.

### Reproduction

Create two implementations of the same flow specification inside a nested component. Give each implementation a distinct incoming connection path. Then reference the nested end-to-end flow from a parent which also has the two corresponding incoming semantic paths.

There are two compatible parent paths, but instantiation currently produces four parent end-to-end flow instances by pairing both incoming paths with both nested variants.

### Expected behavior

An incoming connection instance should be combined only with nested flow instances whose recorded prefix connections match that semantic path. The example should produce two parent flow instances, each connected to the matching nested variant.

### Relevant code

`core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java`, class `ETEInfo` and method `processEndToEndFlow`.

Issue #2985, opened 2026-08-04
Forward-referenced nested end-to-end flows bypass cleanup and clone naming 🤖


lwrage:
### Summary

Forward references to nested end-to-end flows bypass normal cleanup and clone naming during instance-model construction.

When a parent end-to-end flow references a nested flow that has not yet been instantiated, `processEndToEndFlow` creates another `CreateEndToEndFlowsSwitch` and calls `instantiateEndToEndFlow` directly. This bypasses the cleanup performed by `caseComponentInstance`, including removal of incomplete instances and assignment of unique names to cloned instances.

The result depends on declaration order. Feature-group expansion or other branching can leave multiple nested end-to-end flow instances with the same name, and failed nested instances can remain attached to the component instance.

### Reproduction

Declare a parent end-to-end flow before the nested flow that it references. Make the nested flow expand into multiple valid instances, for example by passing through feature-group connections with multiple members.

The nested instances are created through the direct forward-reference path and retain duplicate names instead of receiving the usual numeric suffixes.

### Expected behavior

Nested flow instantiation should have the same cleanup and clone-naming behavior regardless of declaration order. All retained clones should have unique names, and incomplete instances should be removed.

### Relevant code

`core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java`, methods `caseComponentInstance` and `processEndToEndFlow`.

PR #3006, opened 2026-08-07
Make end-to-end flow instantiation atomic 🤖

### LINKED ISSUE #3005: Make end-to-end flow instantiation atomic and maintainable 🤖
## Summary

`CreateEndToEndFlowsSwitch` builds end-to-end flow instances by modifying the instance model during speculative traversal and later undoing unsuccessful changes.

Candidate instances and branch clones are attached to their containing component before validation is complete. Their lifecycle is coordinated through shared mutable fields such as `state`, `connections`, `addETEI`, `removeETEI`, `completedETEI`, `abortedETEI`, `created`, and `myInfo`.

Mode information is also stored temporarily in `EndToEndFlowInstance.getModesList()`, then converted to system operation modes and cleared. Correctness therefore depends on every success, failure, recursion, and cancellation path performing cleanup in the correct order.

This architecture is difficult to understand and maintain. Changes to branching, nested flows, diagnostics, modes, or cancellation can easily leave partially constructed instances attached, discard valid branches, retain invalid branches, or leak transient state between branches.

This issue concerns restructuring the implementation. It does not assert a single new user-visible defect.

## Reproduction

Use an end-to-end flow containing:

- Multiple compatible connection instances or flow implementations.
- A nested end-to-end flow.
- Modal flow elements.
- At least one valid branch and one branch that cannot be completed.

During instantiation, the implementation:

1. Adds the initial `EndToEndFlowInstance` to the component before traversal completes.
2. Uses `EcoreUtil.copy` to create speculative branches.
3. Attaches those clones while other branches are still being evaluated.
4. Mutates shared traversal stacks and connection filters recursively.
5. Records unsuccessful instances in cleanup collections.
6. Renames surviving instances after invalid branches have been removed.
7. Stores mode constraints temporarily on the instance and clears them during finalization.

Cancellation or an exception during this process requires all transient instance-model changes and bookkeeping state to be reversed consistently.

## Expected behavior

End-to-end flow instantiation should:

- Construct candidates independently of the published instance model.
- Keep traversal state local to each branch.
- Validate candidates and finalize their modes before publication.
- Commit completed candidates atomically for each component.
- Leave the original instance model unchanged when instantiation fails or is canceled.
- Preserve deterministic branch ordering, naming, diagnostics, and nested-flow references.
- Avoid using instance-model properties as temporary traversal storage.
- Make candidate status, rollback, and diagnostic ownership explicit.

Existing valid end-to-end flow behavior and protected API contracts should remain unchanged.

## Relevant code

`core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java`

Relevant areas include:

- `caseComponentInstance`
- `instantiateEndToEndFlow`
- `processSubcomponentFlow`
- `processFlowStep`
- `processAccess`
- `processEndToEndFlow`
- `continueFlow`
- `fillinModes`
- Iterator-stack cloning and the `addETEI`/`removeETEI` cleanup mechanism

lwrage:
## Summary

- Build end-to-end flow candidates independently of the published instance model.
- Keep traversal data local to each branch and commit completed candidates atomically per component.
- Roll back cleanly on failed expansion, mode-finalization errors, and cancellation.
- Preserve nested-flow references, diagnostics, mode membership, branch ordering, naming, and protected API behavior.
- Add focused AADL fixtures and characterization tests for basic, branching, access, nested, modal, invalid, refinement, and cancellation scenarios.
- Clarify the implementation and apply Java 21 records, pattern matching, enhanced switches, and sequenced-collection operations.

## Testing

```bash
mvn -o -s releng/osate.releng/settings.xml -Plocal \
  -pl :org.osate.aadl2.instantiation,:org.osate.core.tests,:org.osate.core.feature \
  -Dtycho.localArtifacts=default \
  -Dpr.build=true -Dsign=false \
  '-Dtest=*EndToEndFlowInstantiationTest' -DfailIfNoTests=false \
  clean install
```

All 41 selected tests passed.

Closes #3005


lwrage (commented):
Review of the rewrite, against merge base `cbb8a2878f`. Line numbers refer to `core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateEndToEndFlowsSwitch.java` at head `3dc3730`.

## Overall

The structure is sound: discovery on detached candidates, one `addAll` per component, rollback on failure. Replacing the recursive `new CreateEndToEndFlowsSwitch(...)` for nested flows with a memoized per-component context, and moving `connections` and `lastFlowImpl` out of switch fields into per-branch `TraversalState`, both remove real sources of cross-branch interference. `doc/e2e_instantiation.md` matches the code everywhere I checked it.

Two behavior fixes the PR description doesn't mention:

- Nested selection now filters to `COMPLETE` (lines 1103-1105). On master, `ete2info.get(ete)` returned the nested switch's `created` list, which retained `ETEInfo` entries for instances already removed from `ci.getEndToEndFlows()`. A parent could therefore hold a `flowElement` reference to a detached `EndToEndFlowInstance` — a dangling reference in the saved instance model.
- Diagnostics for discarded candidates no longer target a detached `EndToEndFlowInstance`. Master called `error(etei, ...)` on an instance it had just removed from the component.

Findings 1-3 should be settled before merge. Finding 6 should come out of the PR.

## 1. `processAccess` peeks the shared iterator but continues from the per-branch copy

Lines 1038-1054. `iter.next()` advances the iterator object shared by every branch, while `continueFlow` resumes from `branchState.continuations.pop()`, which for the second and later branches is a copy taken at the fork index.

So branch 2 peeks the segment after the one it should. That segment is usually not a `Connection`, so `getConnection` returns null and the entire backwards walk that reconstructs the intervening `ConnectionReference`s is skipped. In `MultipleAccessTop.i`, branch 1 ends up with `connections = [first, from_boundary]` and branch 2 with `[from_boundary]`. Branch 2 still resolves only because `testConnection`'s single-connection direction check happens to disambiguate the two candidate connection instances. With a multi-hop access connection where an intervening reference is needed to discriminate, branch 2 and later would fail while branch 1 succeeds.

The same asymmetry exists on master (`iter.next()` for the peek, `state.pop()` for the continuation), so this is not a regression. But `processEndToEndFlow` gets it right — it uses `continuation` for both the peek and the resume — so `processAccess` is now inconsistent with its own sibling method, and no test covers the case.

## 2. Forks that are never traversed are promoted to `COMPLETE` and committed

Lines 397-401 turn every still-`ACTIVE` candidate into `COMPLETE`. `processAccess` forks at line 1026, before the `if (leaf != null)` guard at line 1024 has been satisfied for the *next* connection. If the last matching connection's destination is not a data or subprogram proxy, the fork created for it is never entered: it keeps a dangling continuation, stays `ACTIVE`, becomes `COMPLETE`, and `commit` attaches a truncated flow and runs `fillinModes` on it.

`ACTIVE` currently conflates "in progress" with "never started". Either add a `started` flag, or create the fork only once traversal commits to entering it.

## 3. `removeLastFlowImplementation(activeState)` pops from whichever candidate happens to be active

Lines 798, 890, and 903-907. The entry is added to `traversal`, but the removal targets `activeState`, which after `continueFlow` returns is generally a different candidate's state — the last fork in the deepest chain.

Master used a global `Stack<FlowImplementation>` with balanced push/pop. This version is unbalanced, and is benign only because the foreign state always belongs to a candidate that has finished traversal. Nothing states or checks that invariant, and the `if (!state.flowImplementations.isEmpty())` guard silently absorbs the mismatch.

Remove from `traversal` instead, and let it throw if the list is empty. Doc section 6.4 writes this as a plain "pop flowImplementations", which hides the asymmetry.

## 4. New `IllegalStateException`s in a user-facing pipeline

Lines 349, 450, 454, 462, 508, 535, 595. `InstantiateModel.performFinalInstantiation` does not catch these (`InstantiateModel.java:599`), so an internal-invariant violation aborts instantiation of the whole model where master would have produced a wrong-but-present flow. Suggest catching in `caseComponentInstance`, reporting an error on the component, and skipping that component's flows.

On the guard at line 349 specifically: `EndToEndFlow` is an `EndToEndFlowElement` but not a `FlowElement`, so it can only appear in an `EndToEndFlowSegment` at the declaration's own level, where `ci == owner`. The guard is safe, but that reasoning should be a comment on the check.

## 5. `compatibilityInfo` is write-only

Lines 248, 354, 405, 435-446. Nested selection reads `expansions`, so the `ETEInfo` list is built — twice per expansion — and never consumed. The only thing keeping it alive is the protected `instantiateEndToEndFlow(ci, ete, HashMap)` signature, which has no callers in this repository.

Either pass `null` on the internal path and document the parameter as legacy, or deprecate the overload and drop `ETEInfo`. The latter is an API-baseline decision; at minimum, a comment should say why the map is retained. `updateCompatibilityInfo` is also called from both `expandEndToEndFlow`'s `finally` and `instantiateEndToEndFlow`, and only the second matters for an external caller passing its own map.

## 6. `OSATE2.launch` changes are unrelated

The diff drops `org.eclipse.lsp4j`, `org.eclipse.lsp4j.jsonrpc`, `org.eclipse.equinox.slf4j`, `jakarta.activation`, and `com.googlecode.javaewah.JavaEWAH`; adds `javaewah` and `slf4j.simple`; and moves `org.osate.core.tests.log4j` from target bundles to workspace bundles. That is local target-platform drift written back into a config every developer shares, and it is unrelated to flow instantiation. Drop it or split it into its own change.

## 7. Failed candidates keep traversing

Line 1346 bails only on `ABORTED`. A candidate marked `FAILED` by a cycle, an unmatched connection, or an incompatible nested flow continues consuming segments and queuing diagnostics, which now surface as additional `<name> could not be instantiated: ...` messages. Master behaved the same way, so this is not a regression; bailing on `FAILED` as well would cut the cascade, at the cost of updating the expected messages in the new tests.

## Tests

**8.** `InvalidEndToEndFlowInstantiationTest.testMissingFlowImplementationRetainsCompletedFlowAndReportsExactError` asserts the exact full message list, including `No connection declaration from feature input of component composite to subcomponents...`, which comes from `CreateConnectionsSwitch` rather than from this code. Any change there breaks a flow test. The other tests in the class filter by prefix; do the same here.

**9.** `AccessEndToEndFlowInstantiationTest.testFlowStartingAtAccessFeatureRetainsCurrentBoundaryBehavior` and `testUnresolvedBoundaryAccessReportsSystemBoundaryErrors` assert `flow.getFlowElements()` is `List.of()`, so an `EndToEndFlowInstance` with zero flow elements is committed. Downstream analyses index `getFlowElements()`. Please file an issue and cite it in the test, so it reads as a pinned defect rather than intended behavior.

**10.** The commit order pinned by `NestedEndToEndFlowInstantiationTest.testNestedAndParentCloneCounts` (`forward_parent_1, forward_nested_1, forward_nested_2, forward_parent_2`) matches master exactly, so it is a correct characterization. It comes from the global candidate sequence rather than a per-declaration one, which is surprising enough to deserve a one-line comment in the test.

## Doc

**11.** The appendix lists #612, #2872, #2985, #2986, #2987, and #2988 as "behaviors the rework addressed", but all of those mechanisms are already on master at the merge base: `isSameOrRefinedConnection`, `activeEndToEndFlows`/`failedEndToEndFlows`, `containsConnectionPath`, `AadlUtil.hasPortComponents`, and `abortedETEI`. What this PR adds is the nested-endpoint check in `isCompatibleNestedConnection` (commit `3dc3730`) and the `COMPLETE` filter noted above. Reword to something like "behaviors this document explains, with the fix location" so the attributions are not read as this PR's scope. If the nested-matching fix closes an issue, add it alongside `Closes #3005`.

## Minor

- Lines 917 and 963: the `etei` parameter is unused in both `isValidContinuation` overloads, and the new Javadoc documents it as "the candidate being filtered". Both methods are package-private, so either drop the parameter or drop the `@param`.

Issue #3052, opened 2026-08-19
A connection pattern with fewer dimensions than its arrays leaves two identical connections 🤖


lwrage:
Blocked by #3037.

## Summary

A `Connection_Pattern` with fewer dimensions than the arrays it applies to leaves two identical
connection instances behind. Structural expansion reports the mismatch four times, falls back to
the first element of each end, and attaches the result twice, which connection validation then
reports as two connections ending at one data port. Thirteen of the sixteen element pairs the
arrays could have are absent.

Two questions: whether a fallback to the first element is wanted at all, and if it is, why it
happens twice.

Found on 2026-08-18 while completing the connection characterization matrix for #3037. Both the
old and the new traversal produce it identically, so it belongs to structural expansion rather
than to traversal.

## Reproduction

Instantiate `TooFewIndices.i` of
`core/org.osate.core.tests/models/issue3037/StructuralFailures.aadl`, which declares
`Connection_Pattern => ((One_To_One))` over two two-dimensional arrays:

```text
Error | Too few indices ...            (four times)
producers[1][1].outp -> consumers[1][1].inp
producers[1][1].outp -> consumers[1][1].inp
Error | More than one connection instance ends at data port ...
```

## Expected behavior

Either one connection instance per element pair the pattern implies, or one diagnostic and no
connection instance. Two identical connections between the first elements is neither, and it
makes a second, misleading error appear from validation.

## Relevant code

- `InstantiateModel.processConnections()` and the pattern handling it calls — where the
  mismatch is reported and the fallback happens.
- `ValidateConnectionsSwitch` — the secondary duplicate-destination error.
- `Issue3037StructuralFailureTest.aPatternWithTooFewIndicesAgrees` pins the current behavior.

Issue #3051, opened 2026-08-19
A connection inside a referenced classifier root crashes instantiation 🤖


lwrage:
Blocked by #3037.

## Summary

A connection declared inside a component classifier that is referenced as a feature's classifier
crashes instantiation with `NullPointerException`.

Such a classifier is instantiated as a further root of the instance resource, beside the system
instance: `InstantiateModel.instantiateFeatureClassifier()` creates it, and it is a
`ComponentInstance` whose container is the resource, so it has neither a system instance nor a
containing component. The connection phase assumes both. Any connection inside that root reaches
`CreateConnectionsSwitch.getComponentModes()` through `fillInModes()` with a null component,
because the container is resolved from the root's system instance, which is null.

Found on 2026-08-18 while writing the referenced-root characterization fixture for #3037. The
crash predates that work and is independent of it: the traversal issue #3037 replaced reached the
same code from its own attachment step and crashed identically. That work therefore excluded a
referenced root *with connections* from characterization, which is why its acceptance criterion
"referenced roots use the merged complete pipeline" cannot be demonstrated for this shape until
this is fixed.

Before #3037 there was a second, earlier crash on the same shape: a connection travelling *up* to
a feature of the referenced root also failed while asking the containing component for its
declarations, since `InstanceUtil.getComponentImplementation(null, ...)` returns null and
`getAllConnections()` then throws. That path was specific to the removed traversal and no longer
applies.

## Reproduction

A data implementation with a `data` subcomponent, used as the classifier of a data port, with a
connection inside it. Either form crashes:

```aadl
	data implementation Element.i
		subcomponents
			part: data Payload;
			reader: data Consumer.i;
		connections
			-- either of these
			inner: data access part -> read;          -- up to the root's own feature
			between: data access part -> reader.datum; -- between its subcomponents
	end Element.i;
```

used as `signal: out data port Element.i;` on any component that is instantiated.

`core/org.osate.core.tests/models/issue3037/ReferencedClassifiers.aadl` is the same model with
those connections removed, and its header comment says so.

## Expected behavior

A referenced classifier root is instantiated with its connections, and its modes and system
operation modes are resolved against the root itself rather than through a system instance it
does not have. If a connection inside such a root cannot be supported, instantiation reports
that rather than throwing.

## Relevant code

- `CreateConnectionsSwitch.fillInModes()` and `getComponentModes()` — the null container.
- `CreateConnectionsSwitch.attachAcrossFirst()` — where the container is resolved.
[...]

Issue #3050, opened 2026-08-19
Report an upward path the level above does not continue, or decide not to 🤖


lwrage:
Blocked by #3037.

## Summary

When a subcomponent has an inward connection to one of its own features and nothing at the level
above routes that feature anywhere, the connection stops inside the model and no connection
instance exists. Connection instantiation used to report that as `Could not continue connection
from <src>  through <dst>. No connection instance created.` (with two spaces, as it was
written), because it was extending that path when it found out.

Issue #3037 replaced that traversal. The new one never enumerates the path at all: the
declaration that carries it up out of the subcomponent is not an across declaration, so nothing
seeds it and no leg reaches it, and there is no candidate to attach a report to. The warning is
gone; it is allowlist entry 8 of that issue. Reproducing it there would mean seeding every
boundary feature of every subcomponent that nothing above continues, purely to report and
discard the result, which is against the design.

The fact it carried may still be worth reporting, and this issue is about deciding that. The
natural place is the declarative model, where the feature and the missing declaration both
exist, rather than a traversal that happens to be extending a candidate — the same question
issue #3037's other deferred report asks, and the answers may well be the same.

## Reproduction

All three implementations the core test corpus has of this shape:

- `core/org.osate.core.tests/models/Issue2318/findTests/findConnectionInstance.aadl :: X.root`
  — a port
- `core/org.osate.core.tests/models/issue2872/Issue2872.aadl :: integration.impl` — a nested
  feature group
- `core/org.osate.core.tests/models/issue3025/Issue3025.aadl :: Top.i` — a port whose name
  resolves to nothing above

Instantiate any of them: the connection instances are what they were, and nothing is reported
about the path that cannot continue. Before #3037 each produced the warning quoted above.

## Expected behavior

A decision, and if the answer is that this is worth reporting, a report from wherever the fact
is established without a traversal candidate, with a test that pins the target and the wording.

Deciding not to report it is an acceptable outcome;
`Issue3037FailureDiagnosticTest.anUpwardPathTheLevelAboveDoesNotContinueLosesOnlyTheSourceFirstWarning`
already records the three models, and this issue closes as the record of why.

## Relevant code

- `Aadl2JavaValidator` and the declarative connection checks — where a rule about a declaration
  that goes nowhere would live.
- `ValidateConnectionsSwitch` — the alternative, if the fact is better expressed over the
  instance model.
- `core/org.osate.aadl2.instantiation/doc/connection_instantiation.md` §6.

Issue #3049, opened 2026-08-19
Replicated connection references name the wrong array element 🤖


lwrage:
Blocked by #3037.

## Summary

Structural expansion leaves the `connectionReferences` of a replicated connection instance
pointing at the wrong array element. Expansion re-resolves a replica's reference endpoints from
the declaration rather than carrying over the replica's own endpoints, so for a flat feature
group the source end of the chain names the group rather than the member, and for a nested
feature group it also names array element 1 whatever element the replica belongs to. Following
the chain of the element-2 connection therefore walks into element 1.

The connection instance's own `source` and `destination` are correct; only the reference chain
is wrong. Anything that reads the chain instead of the endpoints sees the wrong element: EMV2
propagation paths, flow analysis, and the graphical editor all do.

Found while characterizing connection traversal for #3037, which is why that work compares
structural invariant violations between traversals rather than requiring none: both traversals
produce this identically.

## Reproduction

Instantiate `NestedGroupArrays.i` of
`core/org.osate.core.tests/models/issue3037/ArraysAndFeatureGroups.aadl` and follow the
references of `consumers[2].nest.inner.ack --> producers[2].nest.inner.ack`. Its first
reference has source `consumers[1].nest.inner.ack`, not `consumers[2].nest.inner.ack`.

`Patterned.i`, `Top.nestedGroupArrays` and `Top.patterned` in the same file have it too, four
violations for the patterned pair and two for the plain one.

## Expected behavior

Every connection reference of a replica names the instance objects of that replica, so that the
chain is continuous from the connection's source to its destination. `InstanceIntegrity.check`
in `core/org.osate.core.tests/src/org/osate/core/tests/instantiation/` already expresses the
invariant: a connection's source matches its first reference's source and its destination
matches its last reference's destination.

## Relevant code

- `InstantiateModel.processConnections()` and the replication it performs.
- `Issue3037ArrayFeatureGroupTest.arraysNestedGroupsAndAPatternAgree` and
  `.nestedFeatureGroupsBetweenArraysAgree` pass the current violation counts to
  `InstanceCharacterization.assertConnectionsWithStaleArrayReferences`, so a fix shows up as a
  failure there.

PR #3053, opened 2026-08-19
Use across-first connection traversal 🤖

### LINKED ISSUE #3037: Use across-first connection traversal 🤖
## Summary

Replace source-first semantic connection traversal in `org.osate.aadl2.instantiation` with an across-first, pivot-centered two-leg traversal.

Today `CreateConnectionsSwitch` starts at each candidate source feature of each component and grows a partial path forward through `ConnectionInfo.appendSegment()`, using shared mutable direction and feature-group state (`upFeature`, `downFeature`, `srcToMatch`, `dstToMatch`) and a stack-balancing discipline to decide whether a partial path may still go up, across, or down. The containing component of a semantic connection is only discovered once an across segment happens to be reached, and the source feature path is reconstructed by walking previously collected segments.

An AADL semantic connection has a constrained topology: it travels up zero or more containment levels, crosses between peers exactly once, and travels down zero or more containment levels.

```
ultimate source <- source leg <- across segment -> destination leg -> ultimate destination
```

Seeding traversal at the across declaration exploits that shape directly:

- The across segment identifies the containing component instance immediately.
- Both endpoint feature paths are available up front instead of reconstructed.
- Both legs become one monotonic downward traversal parameterized by role, so `upFeature`/`downFeature` and the balancing logic become independent per-leg state.
- Feature-group endpoint pairs are resolved at the pivot rather than balanced after the destination is reached.
- Leg results are memoizable.

Motivation is maintainability of the connection instantiation code, which has been the source of a long run of defects (issues #3017, #3019, #3021, #3023, #3025, #3027, #3028, #3030, #3032, #3034). It is not a performance enhancement; performance must merely not regress beyond the stated budgets.

## Design

**Seeds.** `Connection.isAcross()` (`ConnectionImpl.java:879-887`) is a purely declarative predicate over source and destination contexts, and includes access connections whose endpoint is a subcomponent. Index declarations by across seed, boundary endpoint and direction, containing component instance, endpoint feature paths, and leg continuation role.

**Degenerate seeds.** Connections with no real across segment get explicit seeds rather than special cases: system input is a boundary seed plus a destination leg, system output is a source leg plus a boundary seed, and other incomplete connections use a one-leg seed.

**Legs.** One `resolveLeg(seedEndpoint, role, pathState)` operation with `SOURCE_LEG` / `DESTINATION_LEG` roles. The source leg follows declarations opposite semantic direction; the destination leg follows them in semantic direction. Per-leg state holds the current endpoint, ordered segments, feature path, mode constraint, internal reverse-traversal capability, and visited oriented declarations.

**Join.** Join only when feature paths and directions are compatible, building `reverse(source leg) + across + destination leg`. Memoize by endpoint, role, feature path, orientation, and mode context.

**Invariant.** A complete internal semantic connection has exactly one containment turning point, so it contains exactly one across segment. This is an architectural invariant, not an empirical observation. The current implementation would silently overwrite `acrossConnection` and `container` on a second across segment (`ConnectionInfo.java:214-218`); across-first must fail immediately instead.

**What does not change.** Path enumeration produces typed immutable path records and creates no EMF objects. Materialization then feeds the existing `InstantiateModel.finalizeConnections()` pipeline unchanged: provisional connection instances are attached, `Connection_Pattern` and `Connection_Set` are cached on them, arrays/patterns/sets are expanded, and the provisional instances are deleted (`InstantiateModel.java:556-605,1421-1517`, deletion at `:1514`). Structural properties are only available on materialized instances and expansion branches on materialized orientation via `Aadl2InstanceUtil.isOpposite()`, so structural properties constrain final expansion, not the join predicate. Mode/SOM assignment stays in `fillInModes()`, including its existing warning-and-delete behavior for paths active in no SOM.

## Pinned baseline

`0639abd94b67d4514f7a149acdb8e302c1d435ab` — `Merge branch '3028_reject_internal_feature_destinations'`.

Every characterization result, name capture, ordering capture, and performance number is produced from that exact commit. All ten prerequisite fixes are verified reachable from it:

| Prerequisite | Issue | Evidence |
|---|---:|---|
| Resolved endpoint invariant | #3017 | `871571455a` |
| Feature-group endpoint mapping | #3019 | `f280111b5c` |
| Short-access destination index | #3021 | `55eef21bb3` |
| Short-access inverse cleanup | #3023 | `64d53825cc` |
| Mode-transition creation | #3025 | `d930fa4cd2` |
| Nullable destination feature | #3027 | `f96faad980` |
| Missing subcomponent instance | #3030 | `26fab7f83b` |
| Final expansion pipeline | #3032 | `44c36e047e`, `30b0fe3456` (direct commits, no PR) |
| Replicated connection contexts | #3034 | `cd326a40b6` |
| Internal features at destinations | #3028 | `d95cf41469`, `9f136becb8`, `0639abd94b` |

The baseline does not move silently. If PR #3006 (issue #3005) or another connection-pipeline change must be included first, a new exact SHA is selected and the baseline audit and capture are re-run.

## Compatibility contract

The externally observable model of the pinned baseline is preserved:

- `ConnectionInstance.bidirectional` stays `false` for normal instantiation, even when all declarative segments are bidirectional. `setBidirectional` is never called during instantiation today; `ConnectionInfo` computes a `bidirectional` field whose accessor `isBidirectional()` (`ConnectionInfo.java:358`) has no caller. `Issue676Test.xtend:101` asserts `false` directly, and `Aadl2InstanceUtil.getBidirectionalConnectionInstances()` consequently returns empty for every normally instantiated model. Internal traversal may track reverse-traversal capability, but must not copy it to the instance attribute. This is explicitly not being "fixed" here.
- Connection-instance names and separators stay byte-for-byte compatible. Names derive only from container-relative source and destination instance paths (`CreateConnectionsSwitch.java:988-1004`).
- Source and destination orientation stays as at the baseline. Orientation is name-visible and is an input to final expansion, which branches on `Aadl2InstanceUtil.isOpposite(conni)` (`InstantiateModel.java:1469,1503`).
- One connection instance per legal orientation is preserved. `Issue2318Test.BiDirPortSimple_samenames` and `BiDirPortSimple_uniquenames` assert two instances for one bidirectional pair, with reference lists `(conn, conn2)` and `(conn2, conn)`.
- No normal mode-transition connection instances reappear.
- Public API and serialized enum values do not change.

### Duplicate enumeration and the reverse-order check

The reverse-order duplicate check at `CreateConnectionsSwitch.java:958-985` rejects a candidate only when source, destination, and reference count also match, so it does not collapse the two legal endpoint orientations above. Its historical motivation is commonly given as issue #565, but that attribution is **unverified**: `git log -S` over the current path resolves only to `2b1d746344` ("renamed directory to match project name", 2016-04-01), which hides the introducing commit.

Rather than assume, this work captures evidence. Temporary source-first instrumentation immediately before duplicate suppression records, for every candidate: stable endpoint paths, ordered declaration URIs, context paths, per-reference reverse flags, feature paths, and accept/suppress decision, with accepted candidates linked to their final materialized descriptors. Suppressed candidates remain non-EMF observations and are never attached merely for inspection.

- If all suppressed candidates normalize to the same semantic path, across-first enumerates that path once and no survivor-orientation rule is needed.
- If suppressed candidates differ in any field that can affect a materialized descriptor, structural expansion, validation, or flow membership, implementation **stops before Phase 3** and the captured candidates are presented for an explicit selection. No canonical winner is invented here.

The `issue565` inverse-feature-group model is added as a characterization fixture for exactly this purpose.

### Approved collection-ordering changes

Across-first enumerates across seeds per container instead of features per component, so insertion order changes. The approved behavior is a deterministic across-first order keyed on containing component instance path, declarative seed URI, legal orientation, endpoint paths, and ordered resolved-segment keys — never object identity or hash iteration order.

Two field-level differences are approved up front, and no others:

1. Per-container `ComponentInstance.connectionInstances` collection order may change from source-first insertion order to the deterministic across-first order.
2. Per-container sibling `endToEndFlows` collection order may change as a consequence. The ordered element sequence *inside* each flow may not change.

Consequently, tests must not use collection position as identity. At the baseline, ten test files index `connectionInstances` positionally, and 25 test files reference `endToEndFlows` — 24 in `org.osate.core.tests` plus one in `analyses/org.osate.modelstats.tests`. Those are converted to structured-identity lookup in a preparatory commit that preserves every assertion about count, endpoints, references, reverse flags, names, properties, and diagnostics.

Ten `.aaxl2` files are checked in (`models/Issue2318/DirectionTests/SameNames/instances/`, `models/issue2780/instances/`, `emv2/.../ConnectionPaths/instances/`). No test loads them by path — they are committed instantiation output, not expected-output baselines — so they are not a comparison gate, but they become stale with respect to the new order.

### Not allowed differences

`bidirectional` becoming true; any name change; source/destination swapping; connection instances disappearing because two orientations were collapsed; connections disappearing from flows; new mode-transition connection instances; changed contextual or modal structural-property results.

Differences are not discovered during differential verification and retroactively labelled intentional.

## Public API decision

`CreateConnectionsSwitch` is `public` in the exported package `org.osate.aadl2.instantiation`, although its only caller is `InstantiateModel.java:563`. It is **retained** as the public compatibility entry point with replaced internals; it is not removed or emptied. `ConnectionInfo` is package-private, so removing it is not an API change.

Note that this repository has **no runnable API baseline check**: the `tycho-p2-extras-plugin` `compare-version-with-baselines` execution is commented out in `releng/org.osate.build.main/pom.xml:471-490` and compares against the remote `lastStableRepository`, not against 2.18.0. The API obligation is therefore satisfied by inspection — `Export-Package` must still list exactly `org.osate.aadl2.instantiation` once the temporary friend-only export is removed, and the public signature of `CreateConnectionsSwitch` must be unchanged — and the before/after signature is recorded in the PR. Enabling a real automated baseline check is separate work.

## Test access and materialized comparison

Internal path types are **not** exposed to `org.osate.core.tests`, and no attempt is made to adapt source-first traversal into a path list it never constructs. Complete materialized instance models are compared instead.

Two new packages:

```
org.osate.aadl2.instantiation.internal    (not exported)
org.osate.aadl2.instantiation.testing      (friend-only export, temporary)
```

```
Export-Package: org.osate.aadl2.instantiation,
 org.osate.aadl2.instantiation.testing;x-friends:="org.osate.core.tests"
```

`org.osate.core.tests` already has `Require-Bundle: org.osate.aadl2.instantiation`, so no new bundle dependency is needed. The internal package holds the traversal, path types, strategy selector, and instrumentation; its members may be `public` for cross-package use because OSGi hides the unexported package from other bundles.

The facade exposes only a narrow API that runs one complete isolated instantiation under a selected strategy and returns the normal materialized instance model and diagnostics. While the duplicate-candidate gate is active it may additionally return immutable normalized candidate observations. It never exposes `ConnectionInfo` or an internal path type. The facade, the friend export, and the dual-strategy tests are removed when source-first traversal is removed; materialized characterization tests against the production path are retained.

Mechanics that matter:

- Strategy selection is constructor or field state on `InstantiateModel`, never static mutable state — there is one construction site (`InstantiateModel.java:563`) and the language server instantiates models concurrently.
- The facade must not use `InstantiateModel.instantiate()` for repeated runs; it creates the instance resource at `getInstanceModelURI(ci)` inside the declarative model's own resource set (`InstantiateModel.java:287-296`), so a second run on the same implementation collides on that URI. Each run reloads the declarative model into a fresh `ResourceSetImpl`, as `buildInstanceModelFile()` does (`:271-280`).
- Diagnostics use a queuing in-memory `AnalysisErrorReporterManager`, not workspace markers.

Because the two runs use different `ResourceSet` objects, EMF identity is never a comparison key. Declarative elements normalize to `EcoreUtil.getURI()`; instance objects to full instance path plus category and array indices; connections to container key, endpoint keys, ordered declarative URIs, ordered context keys, and per-reference reverse flags; diagnostics to severity, exact message, normalized target key, and resource URI; properties to definition URI, normalized applies-to context, modes, and recursively normalized values; flows to containing component key, declarative flow URI, name, modes/SOMs, and ordered normalized element sequence.

Display names are **not** unique — two distinct semantic paths with the same endpoint pair get the same name — so descriptors are keyed on structured identity and names are compared as a multiset under that key, never used as a lookup key. Connection and sibling flow collections are compared as keyed multisets; ordered connection references and ordered elements within each flow are compared exactly.

## Performance fixtures, budgets, noise floor, and abort criteria

Instrumentation in the internal package brackets the connection phase: start immediately before path enumeration, include structural expansion, final materialization, mode/SOM assignment, deletion of inactive connections, and connection validation, stop before end-to-end flow creation. Because expansion, modes, and validation are shared Plan 1 code that both strategies pay equally, a **traversal-only** sub-measurement is also recorded — a phase-level ratio can pass while traversal regresses badly.

One comparable traversal state is defined as an attempt to extend a partial semantic connection by one oriented declarative segment, incremented at source-first `appendSegment()` entry and at the equivalent across-first leg extension boundary. Leg joins are counted separately.

Fixtures:

1. `core/org.osate.core.tests/models/Issue2362/DeclarativeTests.aadl` (existing, ~1400 lines, exercised by `Issue2362InstanceTest`).
2. `AcrossFirstBranchingStress.aadl` — three containment levels, eight compatible source-leg branches, eight compatible destination-leg branches, 64 expected joined paths.
3. `AcrossFirstDepthWidthStress.aadl` — six containment levels, three nested feature-group levels of width four, four source and four destination branches, with an independently asserted expected final-path count.

Budgets:

- `DeclarativeTests.aadl`: median connection-phase time ≤ 1.25× baseline.
- `AcrossFirstBranchingStress.aadl`: median ≤ 1.50× source-first.
- Traversal-only median ≤ 1.50× the source-first traversal-only baseline on every fixture.
- Allocated bytes ≤ 1.50× baseline, or baseline + 128 MiB, whichever is larger.
- Partial-leg-state count: no guessed multiplier. The budget is derived and frozen from a measured source-first calibration curve at branch widths 2, 4, and 8 plus the independently computed final-path count per stress fixture, and must cover both the branching-only and depth-by-feature-group fixtures. The formula and numeric thresholds are recorded in this issue before Phase 3 implementation.

Measurement: two warmup runs, five measured runs, fixed JVM settings, median plus observed spread reported in absolute units — ratios alone are not reportable. Allocation via `com.sun.management.ThreadMXBean.getThreadAllocatedBytes()` after checking `isThreadAllocatedMemorySupported()` and verifying measurement is enabled; `com.sun.management` is not exported by the Eclipse system bundle by default, so it must be reached via `Import-Package` or `org.osgi.framework.system.packages.extra`.

Noise floor: each fixture requires a median measured phase of at least 50 ms, or repeated instantiation inside one measured run until that threshold is met, divided out. If the spread across the five measured runs exceeds the budget margin, the measurement is inconclusive and is repeated with more inner repetitions — not reported as a pass.

**Where measurements run:** timing and allocation budgets are *not* enforced by an ordinary `tycho-surefire` test. Ratios of 1.25/1.50, a 50 ms floor, and "no other load" are not satisfiable on a shared CI agent, and a gated timing test would fail intermittently for reasons unrelated to traversal. They run as an explicitly selected local harness on an idle machine and are recorded here as reviewer-enforced release criteria. The deterministic, machine-independent counters — traversal states, join candidates, final paths, indexed declarations, duplicate candidates, and the frozen state-count formula — remain real gated assertions.

**Abort criterion.** One focused optimization pass means one bounded round of work on seed indexing, leg memoization, or join compatibility, recorded here before it starts. If any budget is still exceeded afterwards, the production default is not switched, the PR is not marked ready, and the failed measurements are recorded here. Closing this issue without merging is an acceptable outcome.

## Coverage gate

For the new traversal package and every materially changed compatibility or materialization method: at least 90% line and 85% branch coverage, treated as floors rather than targets. Every uncovered line and branch is classified as unreachable defensive code, platform-specific, intentionally delegated to an already tested component, or missing test — a missing-test classification fails the gate. Direct behavioral assertions, not mere execution, are required for across/boundary seed classification, both legal orientations, duplicate candidate handling, direct and reached-into feature groups, nested/inverse/subset/empty/partial feature groups, component and feature arrays, structural expansion, mode/SOM compatibility and inactive-path deletion, failure cleanup, property preservation, and flow membership.

For context, current legacy coverage is evidence rather than a gate: `CreateConnectionsSwitch` 84.5% line / 78.5% branch, `ConnectionInfo` 98.8% / 94.4%, `InstantiateModel` 74.0% / 68.3%.

Coverage supplements but does not replace materialized differential comparison, integrity assertions, downstream tests, and performance measurement.

## Relevant code

- `core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/CreateConnectionsSwitch.java` — source-first traversal, duplicate suppression at `:958-985`, name construction at `:988-1004`, dead `duplicate` local and guard.
- `core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/ConnectionInfo.java` — `appendSegment()`, shared `upFeature`/`downFeature`/`srcToMatch`/`dstToMatch` state, `bidirectional` accumulation at `:76,143,153,159`, across overwrite at `:214-218`, unused `isBidirectional()` at `:358`.
- `core/org.osate.aadl2.instantiation/src/org/osate/aadl2/instantiation/InstantiateModel.java` — `finalizeConnections()` at `:556-605`, traversal construction site at `:563`, expansion and orientation branching at `:1421-1517` including `:1469,1503` and provisional deletion at `:1514`, resource-set handling at `:271-296`.
- `core/org.osate.aadl2/src/org/osate/aadl2/impl/ConnectionImpl.java:879-887` — `isAcross()`.
- `core/org.osate.aadl2/src/org/osate/aadl2/util/Aadl2InstanceUtil.java` — `isOpposite()`, `getBidirectionalConnectionInstances()` at `:282-291`.
- `core/org.osate.core.tests/src/org/osate/core/tests/issues/Issue676Test.xtend:101`, `Issue2318Test.xtend` — baseline bidirectional and dual-orientation assertions.
- `releng/org.osate.build.main/pom.xml:471-490` — commented-out baseline comparison.

Working design notes: `codex_plan_2_across_first.md` and `across_first_traversal.md`.

lwrage:
Fixes #3037.

Connection instantiation now enumerates connection instances **across-first**: from each across
declaration, resolving a source leg and a destination leg downward from that pivot, instead of
growing partial paths forward from every candidate source feature. The traversal it replaces is
removed, along with the seam that selected between them.

Baseline: `2b5744966cd30143eaa059b81a4d9de87239c2cf`, the merge of PR #3045 (issue #3044). Every
result below was produced against that commit.

Documentation: `core/org.osate.aadl2.instantiation/doc/connection_instantiation.md` describes the
design and its vocabulary — seed, leg, pivot, path, footprint — and
`org.osate.aadl2.instantiation.internal/package-info.java` carries the same summary in the code.

---

## Design

A connection instance has exactly one containment turning point: it travels up zero or more
containment levels, crosses between peers once, and travels down zero or more levels. The
traversal is built around that shape rather than discovering it.

```text
    ultimate source                                        ultimate destination
          |   <-- source leg --      pivot      -- destination leg -->   |
      c.inner.p  <---- c1 ----  c.p  ==== c2 ====  d.q  ---- c3 ---->  d.inner.q
```

`SeedDiscovery` finds the pivots, plus boundary and mode-transition-trigger seeds for connections
with no crossing of their own. `LegResolver` resolves both legs with one operation, `PathAssembler`
joins compatible legs, `LeafExpansion` narrows the ends to the leaves that become connection
instances, and `PathMaterializer` builds each one. Nothing in EMF exists until materialization, so
a candidate can be explored and discarded without leaving anything behind. Everything except
`CreateConnectionsSwitch` lives in an unexported package.

## Compatibility

Preserved: connection instance names byte-for-byte, endpoint orientation, ordered references with
their contexts and reverse flags, one connection instance per legal endpoint orientation,
`ConnectionInstance.bidirectional == false`, no mode-transition connection instances, and the
merged final pipeline for structural expansion, referenced roots and flow creation.

`CreateConnectionsSwitch` remains the public compatibility entry point with its constructor and
inherited entry point unchanged. `Export-Package` is exactly `org.osate.aadl2.instantiation`, as it
was before the temporary friend-only export was added. This repository has no runnable API baseline
check — the `tycho-p2-extras-plugin` comparison is commented out and targets a remote stable
repository — so the API obligation is met by inspection, recorded here.

Public members removed from `CreateConnectionsSwitch`: `filterIngoingConnections`,
`filterOutgoingConnections` (three overloads), `isDestination`, `isConnectionEnd`,
`filterStartingConnections`. None has a caller anywhere in this repository or in the sibling
`aadl-language-server`.

## Intended differences, approved up front

Eight live allowlist entries. Entry 4 was approved and withdrawn; its number is not reused.

| # | Difference |
|---|---|
| 1 | Per-container `connectionInstances` collection order is a deterministic across-first order rather than source-first insertion order |
| 2 | Per-container sibling `endToEndFlows` order may change as a consequence; the ordered element sequence inside a flow does not |
| 3 | The instantiator warning for a connection whose destination is an internal feature disappears |
| 5 | The instantiator warning that no declaration carries a connection further into a component disappears |
| 6 | A feature a declaration names and the component lacks is reported in the resolver's own words, for both ends |
| 7 | A subcomponent a declaration names and the instance model lacks is reported without the `Instantiation error:` prefix |
| 8 | The instantiator warning that an upward path cannot continue at the level above disappears |
| 9 | Where a pivot's end is a feature group member, the connection instances follow the member the declaration names rather than the enclosing group |

**Entry 9 is the one that changes connection instances**, which the plan otherwise forbids. It was
approved on 2026-08-18 against a traced mechanism and a measurement, and the baseline defect it
allows is filed as #3046: source-first resolved a connection end from the first link of its
`ConnectedElement` chain, so `o.ofg.of1` resolved to `o.ofg`. Losing the member cost it two
orientations differently — one died against a mismatched feature group stack with no diagnostic,
the other spent that stack level on the chain link and materialized a cross product pairing members
the model does not connect. A prototype fix to the old traversal produced exactly what across-first
produces, reference for reference, which is what the decision was taken against. The reviewer chose
to record the defect rather than fix it now.

## Results

**Corpus differential.** Both traversals were run over every parseable model under
`core/org.osate.core.tests/models`, comparing materialized connection descriptors, names included,
over every root of the instance resource.

| outcome | count |
|---|---:|
| identical connection instances | 1354 |
| differing connection instances | 0 |
| across-first fails where source-first succeeds | 0 |
| integrity violations across-first adds | 0 |
| fails under both | 1 (the intentionally invalid `Issue2835` fixture) |
| identical connections, differing diagnostics | 18, all approved entries |

Eleven rules closed the gap from 167 differing implementations to none; each is described in the
enhancement issue with the measurement that motivated it.

**Determinism.** Over 1320 implementations, two runs of the same traversal produce identical
connection and end-to-end-flow collection order, with zero differences.

**Names and downstream.** The whole reactor passes: EMV2, fault tree, ALISA, Assure, GE, model
statistics, slicer, textual instance support, and the analyses. Two `Serializer1Test` expectations
list the same connections in the new document order; `Issue2205Test` compares a broadcast group's
sub-results and warnings as sets, since which member is named first is not a property of the
analysis; two flow-latency `.result` fixtures had four `@connectionInstance.N` indices regenerated
with the test's own `generateResults` switch.

**Coverage.** `org.osate.aadl2.instantiation.internal` is at **91.0% line, 85.9% branch** of the
reachable code, above the 90 and 85 the plan asks for, gated by
`releng/scripts/check-across-first-coverage.py`, which classifies every uncovered line and fails
below either floor. Tycho cannot scope a JaCoCo rule to one package of one bundle, which is why the
gate is a script.

**SpotBugs with FindSecBugs**: passes. **Sibling language server**: compiles, all four modules.

**Performance**, medians of five runs after two warmups, timing only so the instrumentation is not
part of the measurement:

| fixture | source-first | across-first | ratio | budget |
|---|---:|---:|---:|---:|
| `DeclarativeTests.aadl`, all 70 implementations | 114.78 ms | 120.61 ms | 1.05 | 1.25 |
| `AcrossFirstDepthWidthStress.aadl` | 25.73 ms | 28.98 ms | 1.13 | 1.50 |
| `AcrossFirstBranchingStress.aadl` width 8 | 1.22 ms | 1.78 ms | 1.46 | 1.50 |

Across-first spends `2w + 2` traversal states at branch width `w` where source-first spends
`w² + 4w`, and its joins are exactly `w²`, the number of connections the shape has. On the depth
fixture it spends 16 states for 1024 connection instances. Allocated bytes were not measured:
`com.sun.management` is not exported by the Eclipse system bundle, and the plan records that as a
blocked gate rather than something to work around.

## Tests

New characterization, all of it over the production traversal: enumeration and materialization,
inverse and nested feature groups, subset matching, boundary connections, arrays meeting feature
groups, modes and SOMs and sibling flows, all five access kinds and the connection-ending
categories, feature arrays, referenced classifier roots, subprogram call parameters, structural
expansion including `Connection_Pattern` and `Connection_Set`, expansion failures, and failure
diagnostics. 99 implementations across 12 classes record the connection instances they produce.

19 existing test files that used collection position as connection or flow identity now locate
them by structured identity — name, endpoint paths, ordered declarations traversed, reference chain
length, or that the container holds exactly one. Position inside a reference chain and inside a
flow's element sequence is untouched, because that is semantic.

Ten `.aaxl2` files checked in under `models/` are committed instantiation output, not comparison
gates — no test loads them by path — and they become stale with respect to the new collection
order. They are deliberately not regenerated here.

## Legacy removal

`CreateConnectionsSwitch` goes from 2194 to 387 lines. Removed: `appendSegment()`,
`instantiateConnections()`, `instantiateExternalConnections()`, `processIncomingFeature()`,
`balanceFeatureGroupEnds()`, `expandFeatureGroupConnection()`, `finalizeConnectionInstance()`,
`addConnectionInstance()`, the shared `upFeature`/`downFeature` stacks, `FeatureInfo`, and
`ConnectionInfo` — including the reverse-order duplicate comparison and its dead `duplicate` local,
which the census found firing on none of 1246 implementations. Structured path identity
deduplicates before expansion instead, and attachment throws on a duplicate rather than absorbing
it.

The strategy seam, the friend-only testing facade and its `x-friends` export, and the traversal
instrumentation are gone. The frozen state-count curve those counters gated is recorded above
rather than asserted; the five tests that could only compare strategies or read internal keys are
deleted, and coverage was measured to be unaffected by their removal before deleting them.

The last commit modernizes only the files this work created or updated for Java 21: pattern
switches over the sealed seed hierarchy, `var` on about 150 locals, and sequenced-collection
methods. No behavior change.

## Defects found and split out

Each was fixed on its own branch with its own regression test rather than absorbed here, and each
moved the baseline:

- #3038, a crash instantiating a nested boundary feature group member
- #3040, a nested boundary feature group losing the inward connection instance
- #3042, moving the direction check onto the connection instance (reviewer requested)
- #3044, a port sharing a feature group with a connected access feature getting no connection
  instance at all
- #3046, filed rather than fixed, and carried here as allowlist entry 9

## Follow-on work

Raised now that the traversal is the production strategy and the scaffolding is gone, each blocked
by #3037: #3047, #3048, #3049, #3050, #3051, #3052.

One further item is deliberately not filed: `ValidateConnectionsSwitch.checkInDataPortConnections()`
reports duplicate-destination errors while iterating a `HashMap` and a `HashSet`, so their *sequence*
varies between JVM runs on about 40 implementations. It affects both traversals identically and
predates this work; it is why comparing diagnostics as a sequence cannot be a gate at all. Say the
word and I will file it.

---

Do not merge this yet. Please review the evidence above, and entry 9 in particular.


lwrage (commented):
PathKeys.declarative() discards the resource path and keeps only the filename and EMF fragment. As a result, declarations such as `/project-a/model.aadl#/0/@ownedPublicSection/...` and `/project-b/model.aadl#/0/@ownedPublicSection/...` receive identical keys.

These are operational keys rather than report-only labels: `LegResolver.orientedKey()` uses them for cycle suppression, and path assembly uses them for deduplication. Inherited declarations from same-named resources can therefore cause a legitimate continuation or path to be skipped. Please preserve a unique resource or package identity while normalizing only the location-dependent prefix.

lwrage:
> PathKeys.declarative() discards the resource path and keeps only the filename and EMF fragment.

Fixed in ff7cef5187.

`PathKeys.declarative()` now keys a declarative element by the qualified name of its resource root — the `AadlPackage` or `PropertySet` — paired with the file name, then the fragment. So `/project-a/model.aadl#/0/@ownedPublicSection/...` and `/project-b/model.aadl#/0/@ownedPublicSection/...` now differ, both in `LegResolver.orientedKey()` and in `SemanticConnectionKey`. The leading URI segments still stay out of the key, since they are what differ between runs that load the same model from different locations, which is what the normalization was for; the root's qualified name supplies the identity the file name alone does not, and is itself location-independent. A root with no name is not an AADL package, and keeps its whole location, which is unique by construction.

Verified: `Issue3037*` passes, 68 tests, and the whole `org.osate.core.tests` suite passes, 645 tests, no failures.

One thing to flag: I could not construct a model where the old key is observably wrong, so this commit carries no regression test. `orientedKey()` also includes the context instance path, and a descending leg enters a new subcomponent at every hop, so the same context does not recur on one branch; a full `SemanticConnectionKey` collision additionally requires identical endpoints and contexts. I am treating the change as removing a latent hazard rather than fixing a reproduced failure. If you would rather have a fixture attempted — two projects with the same file name, one implementation extending the other's so an inherited and a local declaration share a context — say the word and I will try to make it fail against the old key first.

lwrage:
There's a terminology issue: according to the AADL standard, a semantic connection is a complete connection instance, e.g., from a thread to a thread. Here we create connection instances, which may or may not be complete. I suggest we change the naming.

lwrage:
> There's a terminology issue: according to the AADL standard, a semantic connection is a complete connection instance

Agreed, and fixed in 6ff7bf774a.

`SemanticConnectionPath` is now `ConnectionInstancePath` and `SemanticConnectionKey` is now `ConnectionInstanceKey`. Prose says *connection instance* for the thing produced and *path* for the route it is enumerated from, which the vocabulary already defined. That covers 76 comment and message occurrences across 14 files, the two type names, and 4 test comments.

`doc/connection_instantiation.md` §1 now says why the standard's word is absent, rather than claiming — wrongly — that none of the traversal's vocabulary appears in the standard:

> One word is deliberately absent. The AADL standard's *semantic connection* is the complete case only: a connection between two connection-ending components, thread to thread for instance. What this traversal enumerates and what `CreateConnectionsSwitch` materializes is a `ConnectionInstance`, which may be complete or incomplete — a boundary connection at a feature of the instantiation root is neither thread-to-thread nor invalid.

The vocabulary entry is now **connection instance**, and it names the complete/incomplete distinction explicitly. The PR description above is reworded to match.

Scope: both renamed types live in the unexported `internal` package, so nothing outside the traversal names them, in this repository or in the sibling language server. Uses of the term in code this work did not touch — `ValidateConnectionsSwitch`, `SCProperties`, `InstantiateModel`, the `org.osate.help` documents, the metamodel javadoc — are left alone. Say the word if you want those swept too, though some of them do mean the complete case.

Verified: `org.osate.core.tests` passes, 645 tests, no failures. The coverage gate needs no change; it reads class names from the JaCoCo report rather than a list.

Issue #1833, opened 2019-06-06
Array ports and connections do not appear in flow instances


stevevestal:
## Summary
When I create array ports, connect them, and use those in a flow, the instantiated flow has only source and sink and not the connections.

## Expected and Current Behavior
The instance file does contain multiple connections when arrays are used.  I would have expected the flow to be instantiated in a similar manner.

## Steps to Reproduce
See pasted model and comments.

```aadl
package ArrayFlows
public
	
	data MyType
	end MyType;
	
	process MyProc
		features
			-- If you comment out the [2]
			-- below and instantiate, a
			-- description of the flow will
			-- appear.  Otherwise only a
			-- source and sink appear in
			-- the flow instance.
			inPort: in data port myType[2];
			outPort: out data port myType[2];	
		flows
			outFlow: flow source outPort;
			inFlow: flow sink inPort;
	end MyProc;
	
	system MySys
	end MySys;
	
	system implementation MySys.Impl
		subcomponents
			p1 : process MyProc;
			p2: process MyProc;
		connections
			con: port p1.outPort -> p2.inPort;
		flows
			flw: end to end flow p1.outFlow -> con -> p2.inFlow;
	end MySys.Impl;
	
end ArrayFlows;
```

## Environment

* **OSATE Version**:  2.4.1
* **Operating System**:  Windows 10

AaronGreenhouse:
Okay, I've spent some time reviewing arrays of features and subcomponents in the AADL spec.  I see the section on how connections and `Connection_Pattern` property.

What I don't see is any discussion of arrays and flows.  Arrays are not mentioned in the flow chapter.

I assume what we want in thee above case are two end-to-end flows, one for each connection due to the arrays of features. 

More generally, we need to process all the paths created due to arrays.  This is conceptually not very difficult, but the existing code that creates end-to-end flows is

1. Already extremely confusing, and 
2. Not set up to create more than one `EndToEndFlowInstance` per `EndToEndFlow`

AaronGreenhouse:
Note, we can create the same general scenario as the above example using arrays of subcomponents instead:

```
package ArrayFlows2
public
    
    data MyType
    end MyType;
    
    process MyProc
        features
            inPort: in data port myType;
            outPort: out data port myType;   
        flows
            outFlow: flow source outPort;
            inFlow: flow sink inPort;
    end MyProc;
    
    system MySys
    end MySys;
    
    system implementation MySys.Impl
        subcomponents
[...]

PR #3093, opened 2026-08-25
Expand flow specification instances over feature arrays 🤖

### LINKED ISSUE #2787: Flow specifications need to be able to reference features in a feature array
Properties _Connection_Pattern_ and _Connection_Set_ can be applied to flow specifications. They need to be interpreted during flow instantiation.

This implements saeaadl/aadlv2.2#43

lwrage:
Fixes #2787.

## Cause

`InstantiateModel.instantiateFlowSpecs` creates one `FlowSpecificationInstance` per flow specification and resolves its ends with `findFeatureInstance`, which returns the **first** `FeatureInstance` of a feature array. A flow specification whose in or out end is a feature array therefore became a single instance between element 1 of each end.

`Connection_Pattern` and `Connection_Set` are declared `applies to (connection, flow specification)` in `Communication_Properties`, and were never read there, so there was also no way to say which source element flows to which destination element.

`FlowConnectionMatcher.endsAtFlowSource` then looked the flow specification instance up with `findFlowSpecInstance`, which returns the first one, so even once several instances exist a connection instance reaching any other element matched nothing.

## Correction

`FlowSpecArrayExpander` (new, `instantiation.internal`) replaces each provisional flow specification instance with one instance per pair of elements the properties pair up:

- Both ends arrays of the same size, no property: `One_To_One`, as the property description states for an unset value.
- One end an array, the other scalar: `All_To_One` or `One_To_All`, so every element of the array end has its own flow.
- Any of the 14 `Supported_Connection_Patterns` literals, several alternatives in one value, `Connection_Set`, or a pattern and a set together, whose pairings are the union.
- Size mismatch, an unsupported literal, too few indices, or a pattern on a flow specification with no array end are reported the way the connection expansion reports the same situations, and the provisional instance is kept, so the flow specification is still there for a flow to refer to.

The pairing arithmetic is now `ArrayPatternExpansion`, lifted out of `ConnectionArrayExpander` because a flow specification reads the same two properties with the same meaning; `StructuralProperty` is the one place that identifies those two properties and decodes their values. The connection expander keeps every one of its four messages verbatim.

One instance keeps the name of the declaration; several are named `fs_1`, `fs_2`, … in pairing order, the way the several instances of one end to end flow declaration already are. No metamodel, grammar, property set, or API signature change.

The expansion runs in the window where the two properties are already cached on provisional instances and before the end to end flows are built over them, and it runs for **every** instance model root: it needs only the component's own feature instances, so it does not depend on the system operation modes a referenced classifier root has none of.

End to end flow traversal forks over the instances — at the start of a flow each starts a flow of its own, after a connection the branches are the compatible pairs of connection instance and flow specification instance — and `endsAtFlowSource` takes the instance under consideration instead of looking one up. Nothing else needed an element test: `carriesConnectionPath` already requires the next connection instance to start exactly at the previous element's destination feature instance, so a branch that entered one element stays on it, inside a flow implementation too.

## Regression

`core/org.osate.core.tests/models/issue2787/`, asserted by `Issue2787Test`:

- `Issue2787.aadl` — Aaron Greenhouse's first example in #1833 (a flow source and a flow sink on feature arrays, connected element by element), and the same flow through a component whose flow specification is realized by a flow implementation. Asserts the flow specification instances with their paired elements, and the end to end flow instances: `etef_1` through element 1 and `etef_2` through element 2, with the implementation staying on the element its flow specification instance is for.
- `Issue2787Patterns.aadl` — one flow specification per pairing: the default, `Next`, two alternatives in one value, `Connection_Set`, a pattern and a set together, one array end against one scalar end in both directions, and a flow source and a flow sink whose single end is an array. Asserts every resulting instance with its source and destination array element.
- `Issue2787Failures.aadl` — size mismatch, a pattern with no array end, and a pattern needing an index the scalar end cannot give. Asserts the reports and that the flow specification instance survives.

All five cases fail before the fix with a single instance on element 1, a single flow instance, and no report.

The array ends sit on devices and abstract components rather than on the processes of the original issue, because a feature array is allowed only in an abstract, thread, device, memory, system, or processor classifier.

## Validation

Focused, from the repository root:

    mvn -o -s releng/osate.releng/settings.xml -Plocal -T6 \
      -pl :org.osate.aadl2.instantiation,:org.osate.core.feature,:org.osate.core.tests \
      -Dtycho.localArtifacts=default -Dpr.build=true -Dsign=false \
      -Dtest=Issue2787Test -DfailIfNoTests=false clean install

`Issue2787Test` 5/5. The whole of `org.osate.core.tests` is 800/800, including the 96 tests over connection arrays, patterns and connection sets, which pass unchanged after the extraction.

Clean root reactor:

    mvn -s releng/osate.releng/settings.xml -Plocal -T6 \
      -Dtycho.localArtifacts=ignore -Dpr.build=true -Dsign=false \
      -DfailIfNoTests=false clean install

BUILD SUCCESS, 1395 tests, 0 failures, 0 errors.

## Dependencies

None. Based on `master` at 61e405b81a. #3005, which #2787 was marked blocked by, is merged.

## Residual risk

Instantiation output changes for any model that has a flow specification over feature arrays: n flow specification instances named `fs_1`…`fs_n` instead of one named `fs` on element 1, and correspondingly more end to end flow instances, so flow analyses such as latency will report more flows. That is the point of the issue, but it is a visible change for existing models.

One existing characterization recorded the old behavior and is updated here: `Serializer1Test.testFlowSpecification`, where a flow sink on a feature array now serializes as `f4_1` and `f4_2`. Its two other array cases are unchanged, and the test now says why — an array declared inside a feature group type is not allowed and is instantiated as a single feature, and a feature reached *through* an array feature group is not itself an array feature.

A flow end reached through an array feature group keeps the first element of that group, as before. That is deliberate scope for this change and worth its own issue if it matters.

`Connection_Pattern` and `Connection_Set` on a flow specification are reported only during instantiation, as they are for connections; nothing was added to the editor's validation.

### DIFF
--- a/core/org.osate.aadl2.instance.textual.tests/src/org/osate/aadl2/instance/textual/tests/Serializer1Test.java
+++ b/core/org.osate.aadl2.instance.textual.tests/src/org/osate/aadl2/instance/textual/tests/Serializer1Test.java
@@ -719,6 +719,13 @@ public class Serializer1Test extends AbstractSerializerTest {
 		//@formatter:on
 	}
 
+	/**
+	 * {@code f4} is a flow sink on a feature array, so it has one flow specification instance per array
+	 * element, named {@code f4_1} and {@code f4_2} (issue #2787). {@code f6} and {@code f7} are not
+	 * expanded: {@code p5} is an array declared inside a feature group type, where an array is not allowed
+	 * and is instantiated as a single feature, and {@code f7}'s end is a feature reached through an array
+	 * feature group rather than an array feature itself.
+	 */
 	@Test
 	public void testFlowSpecification() throws Exception {
 		var pkg1 = testHelper.parseString("""
@@ -793,7 +800,8 @@ public class Serializer1Test extends AbstractSerializerTest {
 					flow f1 ( -> p2 ) : pkg1::s:f1
 					flow f2 ( p1 -> ) : pkg1::s:f2
 					flow f3 ( p1 -> p2 ) : pkg1::s:f3
-					flow f4 ( p3[1] -> ) : pkg1::s:f4
+					flow f4_1 ( p3[1] -> ) : pkg1::s:f4
+					flow f4_2 ( p3[2] -> ) : pkg1::s:f4
 					flow f5 ( fg1.p4 -> ) : pkg1::s:f5
 					flow f6 ( fg1.p5 -> ) : pkg1::s:f6
 					flow f7 ( fg2[1].p6 -> ) : pkg1::s:f7
--- a/core/org.osate.aadl2.instantiation/doc/connection_instantiation.md
+++ b/core/org.osate.aadl2.instantiation/doc/connection_instantiation.md
@@ -82,18 +82,20 @@ InstantiateModel.fillSystemInstance(root)
  │ 2. createSystemOperationModes()  → SystemOperationMode list            │
  │ 3. CreateConnectionsSwitch       → ConnectionInstances            ◄── us│
  │      seeds → legs → paths → expansion → materialization → modes/SOMs   │
- │ 4. cacheStructuralConnectionProperties() + processConnections()         │
+ │ 4. cacheStructuralProperties() + processConnections()                   │
  │      → array replication, Connection_Pattern, Connection_Set           │
- │ 5. ValidateConnectionsSwitch     → direction, classifier and           │
+ │ 5. processFlowSpecifications()    → one flow specification instance per │
+ │                                     pair of feature array elements     │
+ │ 6. ValidateConnectionsSwitch     → direction, classifier and           │
  │                                    duplicate-destination checks        │
- │ 6. CreateEndToEndFlowsSwitch     → EndToEndFlowInstances               │
- │ 7. cacheProperties(), annex instantiation                              │
+ │ 7. CreateEndToEndFlowsSwitch     → EndToEndFlowInstances               │
+ │ 8. cacheProperties(), annex instantiation                              │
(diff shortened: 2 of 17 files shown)


lwrage:
Pushed 53c0d6b904, tests only: a nested end to end flow replicated by a feature array.

The nested-flow cases elsewhere take their multiplicity from arrays of subcomponents, which is #1833. This one has no subcomponent array at all — `sensor`, `worker` and `actuator` are single components, and the multiplicity is entirely the flow specification instances this PR creates. `Top.nested` puts the nested flow at the end of its container and `Top.nestedAtStart` at the start, the two handshakes `processEndToEndFlow` has:

    inner_1 : wpath_1 -> worker.outp[1] --> actuator.inp[1] -> fsnk_1
    outer_1 : fsrc_1 -> sensor.outp[1] --> worker.inp[1] -> inner_1

    inner_1 : fsrc_1 -> sensor.outp[1] --> worker.inp[1] -> wpath_1
    outer_1 : inner_1 -> worker.outp[1] --> actuator.inp[1] -> fsnk_1

with `inner_2`/`outer_2` on element 2 in both, and no crossing between elements: each replica of the container refers to the nested replica of its own element. That falls out of `isCompatibleNestedConnection` comparing feature instances, together with each instance of a flow specification over feature arrays having a source feature instance of its own.

Both fail on `master`, each with a single `inner` and a single `outer` on element 1.

`Issue2787Test` 6/6, `org.osate.core.tests` 801/801, clean root reactor BUILD SUCCESS with 1396 tests, 0 failures, 0 errors.

Issue #3032, opened 2026-08-12
Finalize connections before validation and flows 🤖


lwrage:
## Summary

Connection expansion runs at the very end of instantiation, after connection validation and after
end-to-end flow creation. `InstantiateModel.fillSystemInstance()` runs `CreateConnectionsSwitch`,
then `ValidateConnectionsSwitch`, then `CreateEndToEndFlowsSwitch`, and only afterwards
`cacheProperties()` followed by `processConnections()`.

`processConnections()` is what expands subcomponent arrays, `Communication_Properties::Connection_Pattern`
and `Communication_Properties::Connection_Set`. It creates the expanded replacement connection
instances and then deletes the provisional originals with `EcoreUtil.delete()`.

Two observable consequences:

1. `EndToEndFlowInstance.flowElements` is a non-containment reference list. Deleting a provisional
   connection removes it from every end-to-end flow instance that was already created, and nothing
   inserts the expanded replacements. The flow instance silently loses its connection segments.
2. The expanded connections never reach `ValidateConnectionsSwitch`, so instance-level connection
   errors that only exist after expansion are never reported.

The ordering is not a simple call reorder: `processConnections()` reads `Connection_Pattern` and
`Connection_Set` from property associations that `cacheProperties()` cached on the provisional
connection instances, so moving the expansion earlier requires evaluating those two structural
properties before final materialization.

`instantiateFeatureClassifier()` repeats the same ordering for referenced classifier roots, and in
addition instantiates their annexes before any connection expansion or property caching has run for
any root. A fix limited to `fillSystemInstance()` would therefore be incomplete.

### 1. End-to-end flow loses its connection

```aadl
package Flows
public
	thread Producer
	features
		outp: out data port;
	flows
		fsrc: flow source outp;
	end Producer;

	thread Consumer
	features
		inp: in data port;
	flows
		fsnk: flow sink inp;
	end Consumer;

	process Container
	end Container;

	process implementation Container.i
	subcomponents
		producers: thread Producer[2];
		consumers: thread Consumer[2];
	connections
		c: port producers.outp -> consumers.inp;
	flows
		etef: end to end flow producers.fsrc -> c -> consumers.fsnk;
	end Container.i;

	system Top
	end Top;

	system implementation Top.i
	subcomponents
		container: process Container.i;
	end Top.i;
end Flows;
```

The model validates with no issues. Instantiating `Top.i` produces the two expanded connection
instances

```
producers[1].outp --> consumers[1].inp
[...]

PR #3094, opened 2026-08-25
Resolve flow segments to every element of a subcomponent array 🤖

### LINKED ISSUE #1833: Array ports and connections do not appear in flow instances

lwrage:
Fixes #1833.

**Depends on #3093 and is based on its branch, not on `master`.** Merge #3093 first; GitHub will retarget this one to `master` when it does. The diff shown here is this change alone.

## Cause

`EndToEndFlowSession.processETESegment` resolves a segment that names a subcomponent with `ComponentInstance.findSubcomponentInstance`, which returns the **first** component instance of that subcomponent. For an array of subcomponents that is element 1, so an end to end flow over such an array became a single flow instance through element 1, however many connection instances the array connection stood for.

With a pattern that crosses the indices it became no flow instance at all: no connection instance goes from element 1 to element 1, so the traversal reported "there are no semantic connections that connect to the start of the flow" and failed the candidate.

## Correction

Both segment kinds that name a subcomponent resolve to all of its component instances, and `forkOverElements` continues a flow per element. A declaration names a subcomponent, not one of its elements, so all of them are possible continuations.

The elements are **narrowed before** the fork, to those some pending connection instance ends in. Forking into an element no connection reaches would create a branch that can only die, and `processFlowStep` would report that death as a missing semantic connection — so issue 1984's rule that the endpoint predicates are a filter and not an error reporter has to be applied one fork earlier now that a declaration can stand for several elements. When nothing reaches any element, one element is kept, so the report is still made exactly once, as it was when a declaration resolved to the first element only. Narrowing before the fork is also what keeps the flow numbering in element order.

A nested end to end flow replicated this way replicates the flow that contains it **without further change**: `processEndToEndFlow` already forks over the nested expansion's candidates, and `FlowConnectionMatcher.isCompatibleNestedConnection` compares feature instances rather than declarative features, so each replica of the container pairs with the nested replica of its own element. This PR adds the test that pins it.

## Regression

`core/org.osate.core.tests/models/issue1833/Issue1833.aadl`, asserted by `Issue1833Test`, is Aaron Greenhouse's second example in the issue plus four more shapes of the same defect:

- `Top.i` — the example as written. Two flow instances, each through the array element of its own index.
- `Top.patterned` — `One_To_One` written out, same result, since it is the default.
- `Top.crossed` — `Cyclic_Next`. Each flow ends in the *other* element of the receiver array, reachable only if the segment resolves to every element.
- `Top.broadcast` — `All_To_All`, a flow per pair of elements. The numbers and the paths are asserted separately, because which flow gets which number depends on the order the branches were forked in.
- `Top.fanIn` — one array end against one scalar end, `All_To_One`, both emitters flowing into the one receiver.
- `Top.nested` — a nested end to end flow replicated per element, asserting that its container is replicated with it and that `outer_1` refers to `inner_1` and `outer_2` to `inner_2`.

Every element path is asserted by its path relative to the system instance, so which array element a flow goes through is part of what is checked.

All six cases fail before the fix. The `One_To_One`, `All_To_All` and fan-in cases produce one flow instance through element 1; the crossing pattern produces none at all.

## Validation

Focused, from the repository root:

    mvn -o -s releng/osate.releng/settings.xml -Plocal -T6 \
      -pl :org.osate.aadl2.instantiation,:org.osate.core.feature,:org.osate.core.tests \
      -Dtycho.localArtifacts=default -Dpr.build=true -Dsign=false \
      -Dtest=Issue1833Test -DfailIfNoTests=false clean install

`Issue1833Test` 6/6. The whole of `org.osate.core.tests` is 806/806.

Clean root reactor:

    mvn -s releng/osate.releng/settings.xml -Plocal -T6 \
      -Dtycho.localArtifacts=ignore -Dpr.build=true -Dsign=false \
      -DfailIfNoTests=false clean install

BUILD SUCCESS, 1401 tests, 0 failures, 0 errors.

## Residual risk

Instantiation output changes for any model with an end to end flow over an array of subcomponents: one flow instance per element instead of one through element 1, so flow analyses will report more flows. That is the point of the issue.

`Issue3032Test` recorded the old behavior for three of its models and is updated here. Two now have a flow instance per element. The third, `unconnectableEndToEndFlowIsReported`, asserted that a crossed `Connection_Set` produces no flow and reports the missing connection — which *was* this defect — so it is renamed to `aCrossedConnectionSetConnectsTheOtherElement` and now asserts the two flows. What #3032 is about, that a flow is built over the final connection instances and refers to none that was deleted, is still asserted for all three.

A flow over a large subcomponent array forks once per element at its first segment before any connection can narrow the choice, so an array of n elements starts n branches there. Each is pruned at the next hop, and this is the same shape the connection expansion already has, but it is worth knowing for very wide arrays.

### DIFF
--- a/core/org.osate.aadl2.instantiation/doc/e2e_instantiation.md
+++ b/core/org.osate.aadl2.instantiation/doc/e2e_instantiation.md
@@ -123,7 +123,7 @@ end to end flow
   consumer.fsnk
 ```
 
-Five sources of multiplicity:
+Six sources of multiplicity:
 
 1. **Multiple flow implementations** for one flow specification — in the test model
    `BasicAndBranching.aadl`, `Choice.two_paths` declares `fpath` twice, once through `upper` and
@@ -136,6 +136,9 @@ Five sources of multiplicity:
    out end is a feature array was expanded by `FlowSpecArrayExpander` into one instance per pair of
    array elements that `Connection_Pattern`, `Connection_Set` or the default `One_To_One` pairs up
    (issue #2787).
+6. **Multiple component instances** for one subcomponent named by a segment — a subcomponent array has
+   one component instance per element, and each of them continues a flow of its own (issue #1833).
+   A nested ETE replicated this way replicates the flow that contains it, through source 4.
 
 Each multiplicity point **forks** the traversal. Surviving branches are named `<flow>_1`,
 `<flow>_2`, … at commit time.
@@ -331,17 +334,27 @@ processETESegment(ci, etei, segment, iter, errorElement)
   │                            else:
   │                                traversal.connections += fe   (pending filter)
   │
-  ├─ FlowSpecification ──────► sci = ci.findSubcomponentInstance(segment.getContext())
-  │                            sci != null → processSubcomponentFlow(sci, …)
-  │                            sci == null → owner error, "Could not find component
-  │                                          instance for subcomponent …"
+  ├─ FlowSpecification ──────► scis = subcomponentInstances(ci, segment.getContext(), …)
+  │                            FORK PER ELEMENT → processSubcomponentFlow(element, …)
+  │                            scis empty → owner error, "Could not find component
+  │                                         instance for subcomponent …"
   │
-  ├─ Subcomponent ───────────► processFlowStep(sci, …)  (implicit flow: whole component)
+  ├─ Subcomponent ───────────► scis = subcomponentInstances(ci, fe, …)
+  │                            FORK PER ELEMENT → processFlowStep(element, …)
+  │                            (implicit flow: the whole component)
   │
   ├─ DataAccess / SubprogramAccess ─► processAccess(…)
   │
   └─ EndToEndFlow ───────────► processEndToEndFlow(…)   (nested flow)
 ```
+
+A declaration names a subcomponent, not one of its elements, so `subcomponentInstances` returns all
+of them and `forkOverElements` continues a flow per element (issue #1833). It **narrows** them first:
+with pending connections, only the elements some matching connection instance ends in survive.
+Forking into an element no connection reaches would create a branch that can only die, and
+§6.4 would report that death as a missing semantic connection — so issue 1984's filter-first rule
+has to be applied one fork earlier. When nothing reaches any element, one element is kept, so the
+report is made once, the way it was when a declaration resolved to the first element only.
 
 ### 6.2 Descent and ascent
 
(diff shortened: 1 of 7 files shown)


lwrage:
Pushed 270965933b, tests only: the nested end to end flow case now covers all four positions a replicated nested flow can be in, not just the one where it ends its container.

Each takes a different path through `processEndToEndFlow`, so each was worth pinning separately:

| Model | Position | What it exercises |
|---|---|---|
| `Top.nested` | ends the container | the container reaches the nested flow along a connection: `isCompatibleNestedConnection` over `conni × nested` |
| `Top.nestedAtStart` | starts the container | no connection constrains the choice, so the fork is over the nested replicas themselves, and the filter for what follows is seeded from the replica's `postConnections` |
| `Top.nestedInMiddle` | in the middle | both of the above in one flow |
| `Top.nestedThroughImplementation` | starts inside a flow implementation | the nested flow carries a leading declarative connection, so `containsConnectionPath` has to accept the container's connection instance |

In all four the container is replicated per element and each replica pairs with the nested replica of its own element, for example

    inner_1 : firstRelays[1].fpath -> firstRelays[1].outp --> secondRelays[1].inp -> secondRelays[1].fpath
    outer_1 : emitters[1].fsrc -> emitters[1].outp --> firstRelays[1].inp -> inner_1 -> secondRelays[1].outp --> receivers[1].inp -> receivers[1].fsnk

and all four fail without the fork over subcomponent array elements, each with a single `inner` and a single `outer` through element 1. So the replication of a container by its nested flow needs no production change of its own beyond this PR: `processEndToEndFlow` already forks over the nested expansion's candidates, and the pairing is element-exact because `isCompatibleNestedConnection` compares feature instances rather than declarative features.

`Issue1833Test` is 9/9, `org.osate.core.tests` 809/809, clean root reactor BUILD SUCCESS with 1404 tests, 0 failures, 0 errors.

lwrage:
Rebased onto the new head of #3093, which gained a tests-only commit, so this branch stays linear on top of it. No change to this PR's own four commits beyond the new parent.

Clean root reactor after the rebase: BUILD SUCCESS, 1405 tests, 0 failures, 0 errors.

lwrage:
This PR did not reach `master`, so #1833 is not fixed on `master` yet.

It was opened against `2787_flow_spec_feature_array_expansion` so its diff would show this change alone, with the expectation that GitHub would retarget it to `master` once #3093 merged. GitHub retargets an open PR only when its base branch is **deleted**, and that branch still existed, so:

- 20:19:43 — #3093 merged into `master`
- 20:20:11 — this PR merged into `2787_flow_spec_feature_array_expansion`, which had already been merged and was a dead end

The four commits are on `origin/2787_flow_spec_feature_array_expansion` and not on `master`:

    $ git log --oneline origin/master..origin/2787_flow_spec_feature_array_expansion
    769cb853a6 Merge pull request #3094 from osate/1833_subcomponent_array_flow_segments
    d54c4dd7eb Cover every position a replicated nested flow can be in
    405fe6008c Document the fork over subcomponent array elements
    c9b856a5fa Resolve flow segments to every element of a subcomponent array
    01bb6511af Add regression test for issue #1833

Superseded by #3097, which is the same four commits rebased onto current `master` and revalidated there. Nothing in them changed.

`origin/2787_flow_spec_feature_array_expansion` and `origin/1833_subcomponent_array_flow_segments` can be deleted once #3097 merges.

PR #3151, opened 2026-09-03
Avoid duplicate Behavior Annex subclause diagnostics 🤖

### LINKED ISSUE #2445: Port behavior annex implementation to Xtext
## Summary

The behavior annex implementation uses a hand written ANTLR parser. It should be ported to Xtext to benefit from advanced Xtext functionality. It would also allow us to get rid of special handling for non-Xtext annexes in OSATE.

## Environment

* **OSATE Version**: 2.8.0
* **Operating System**: all

lwrage:
Fixes #2445

## Cause

The BA qualified-name provider excluded generated BA model objects from the global index but still delegated the host `DefaultAnnexSubclause` to the Properties name provider. Each embedded BA wrapper was therefore exported as the same global `behavior_specification` element, and Xtext unique-name validation reported false duplicates when a file contained multiple BA subclauses.

## Correction

- Exclude every `AnnexSubclause` from BA global qualified-name export, matching the embedded-subclause boundary used by EMV2.
- Add an embedded AADL regression model with two systems that each contain a BA subclause and assert that validation produces no issues.
- Remove the false duplicate diagnostics from BA characterization goldens and update the corresponding legacy issue counts.

## Validation

- Focused BA Xtext integration reactor: 6 tests, 0 failures, 0 errors.
- Complete BA test bundle: 62 tests, 0 failures, 0 errors, 10 intentional skips.
- Clean root reactor with `-o -T5 -Dtycho.localArtifacts=ignore`: all 143 projects passed.
- `git diff --check` passed.

## Dependency

This PR is stacked on #3150 and should merge after it. It is the closing PR for issue #2445.

## Residual risk

The change is limited to BA subclauses and subclause-local declarations. BA has no annex-library form, and the corpus characterization confirms that unrelated diagnostics and translated models remain unchanged.

### DIFF
--- a/ba/org.osate.ba.tests/expected/diagnostics/org.osate.ba.tests_models_covering_semantic_lr_D3_L1_L2.txt
+++ b/ba/org.osate.ba.tests/expected/diagnostics/org.osate.ba.tests_models_covering_semantic_lr_D3_L1_L2.txt
@@ -1,3 +1,0 @@
-error | semantic | 31 | 7 | 22 | Duplicate Element 'behavior_specification' in SubprogramImplementation 'sub.good'
-error | semantic | 41 | 7 | 22 | Duplicate Element 'behavior_specification' in SubprogramImplementation 'sub.error1'
-error | semantic | 56 | 7 | 22 | Duplicate Element 'behavior_specification' in SubprogramImplementation 'sub.error2'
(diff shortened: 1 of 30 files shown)


lwrage:
merged in 5c5c5f07bfc1bfed779e6c43e370b6f8fce5693f

